PR #563, opened 2017-12-04
Function tests

jvesely:
This PR adds function tests for distance, transfer, and integrator functions.
It also adds dependence on pytest_benchmark, although benchmarks are disabled by default
The two integrator function failures (noise array) can be fixed by the following change:
```
diff --git a/psyneulink/components/functions/function.py b/psyneulink/components/functions/function.py
index fedb76273..9c10dee06 100644
--- a/psyneulink/components/functions/function.py
+++ b/psyneulink/components/functions/function.py
@@ -4026,7 +4026,7 @@ class Integrator(IntegratorFunction):  # ---------------------------------------
             if len(noise) == 1:
                 pass
             # Variable is a list/array
-            elif not iscompatible(np.atleast_2d(noise), var) and len(noise) > 1:
+            elif not iscompatible(np.atleast_1d(noise), var) and len(noise) > 1:
                 raise FunctionError(
                     "Noise parameter ({}) does not match default variable ({}). Noise parameter of {} must be specified "
                     "as a float, a function, or an array of the appropriate shape ({})."
```
It's more of a workaround than a fix, so it's not part of this PR. Feel free to delay merging until a proper fix is in place


buildbot-princeton:

buildbot-princeton:

jvesely:
tho workaround breaks mechanism tests, so definitely not a fix

kmantel:
One thing - even if disabled, the benchmarking packages need to be added to dev_requirements.txt because of the pytest arguments set in setup.cfg.

The problem with noise seems to be a bug in how things are validated relevant to how variables are assigned. Currently, Mechanisms convert their default variable to 2d array, but Functions do not. This should hopefully be taken care of/ I can make sure this is addressed in the refactoring of default variable assignment. For now, a workaround could be checking compatibility for either atleast_2d or atleast_1d

PR #807, opened 2018-05-23
Recurrent input state

dcw3:


dcw3:
updated this to use Katherine's new _parse_function_variable instead of _execute

dcw3:
forgive me for the failed pull requests. the next one should work now!

PR #815, opened 2018-06-20
add macos builds to travis

jvesely:
There might be a more efficient way to do this, rather than install everything via brew, but this is enough to pass the test suite.


jvesely:
The preinstall setup is based on https://pythonhosted.org/CodeChat/.travis.yml.html

PR #821, opened 2018-06-21
Feat/function/relu

karoara:
Adding implementation of the relu activation function to functions.py, test for the function in test_transfer_mechanism.py, and relevant keywords in keywords.py

JDC:
Function.Relu:
  - renamed ReLU 
  - modified function to use np.maximum for elementwise operation on array:
      (conditionals on arrays are not allowed by numpy;
       want elementwise rather than array-wise application of tests and transform)
  - see additional notes below


buildbot-princeton:

buildbot-princeton:

jdcpni:
Problems with benchmark in test - can't be used more than once; 
           changed to simple call of T.execute(<test input>) for now
Test should validate use of bias and leak

PR #975, opened 2018-11-07
Refactor/opt ctl mech/pred vec to curr state

jdcpni:
• OptimizationControlMechanism, LVOCControlMechanism:
  - prediction_vector -> current_state
  - previous_prediction_vector -> previous_state


fredcallaway:
This is a bit late, but I think `state_features` and `previous_state_features` would be more informative names.

PR #986, opened 2018-12-03
Bring PTX/CUDA execution to composition level

jvesely:
Implement buffer handling for mechanism and composition execution classes
add tests


buildbot-princeton:

buildbot-princeton:

PR #1011, opened 2018-12-23
Store compiler structures as compiler parameters

jvesely:
Cleanup llvm and cuda tests. automatically skip cuda tests if cuda is not available.
Use dump files for IR/ISA dumps


buildbot-princeton:

buildbot-princeton:

PR #1013, opened 2018-12-24
Avoid using compilation_data.execution

jvesely:
The param is now CUDA only. It keeps GPU side buffers.


buildbot-princeton:

buildbot-princeton:

PR #1039, opened 2019-01-23
.travis: Don't cache installed python modules

jvesely:
Make sure we always use the latest version.


buildbot-princeton:

PR #1043, opened 2019-01-24
learningfunctions/Kohonen: Correctly report faulty variable in the error message

jvesely:


buildbot-princeton:

PR #1051, opened 2019-02-04
Autodiff clean up

dcw3:


dcw3:
The additions I made seem to reduce the speed of the AutodiffComposition, by a decent amount. In the future, it would be valuable to return to this and optimize it.

PR #1071, opened 2019-02-11
Fix logging none time

dcw3:


dcw3:
@kmantel @KristenManning

PR #1237, opened 2019-06-26
Default keyword change in ModulatorySignal constructor, "projections"/"modulates" equivalence in ModulatorySignal constructor, DDM bugfix

dillontsmith:


psyneulink/library/components/mechanisms/processing/integrator/ddm.py
```diff
@@ -1164,14 +1164,17 @@ def _execute(
                 return_value[self.DECISION_VARIABLE_INDEX] = threshold
             return return_value
 
-    def reinitialize(self, *args, execution_context=None):
+    def reinitialize(self, *args, execution_context=NotImplemented):
```
kmantel: If using most recent execution id is the way we want to go with no-arg reinitialize, we should do it in a way that handles all the reinitialize methods in its own commit/pr

dillontsmith:
> If using most recent execution id is the way we want to go with no-arg reinitialize, we should do it in a way that handles all the reinitialize methods in its own commit/pr

Handling all the reinitialize methods in one commit/pr makes sense. I'll revise and in the meantime solicit some feedback on whether most recent execution id is the right way forward

PR #1240, opened 2019-06-28
Consistent docstrings

jvesely:
Consistently use triple double quotes for docstrings (PEP 257).
Disable typehints in the documentation.


jvesely:
@jdcpni, @kmantel, this creates a lot of noise. I've checked that the output of
`sphinx-build -b html -aE docs/source` is almost the same with the only exception of removing type hints. Type hints produced output like:
`<em class="sig-param">condition: &lt;typecheck.framework.optional object at 0x7f730808cba8&gt; = None</em>`
but also:
`<em class="sig-param">noise: psyneulink.core.globals.utilities.parameter_spec = 0.0</em>`

It looks like typecheck could use some pretty-printing, so I disabled it ftm.
It might be fixed in later version (pnl is restricted to typecheck==1.2)
I've also some work in progress to use raw strings and avoid the need for extra escaping (backslashes), but I wanted to get this in early to avoid further conflicts.

kmantel:
For the quotes, I don't really see it being any better to change it in the future, so now is as good a time as any. For the type hints, I'm having trouble seeing when/why what you describe is happening. I see one example for `prefs` on `WaldDist` that's as you describe, but on the devel docs it's showing up like this:
`class psyneulink.core.components.functions.distributionfunctions.WaldDist(default_variable=None, scale=1.0, mean=1.0, params=None, owner=None, prefs: <function is_pref_set at 0x109d6c0d0> = None)`
which is weird looking. I haven't been able to find the other examples though (noise, condition).

jvesely:
hm, that's interesting. I saw the type hint cases mostly in transfer functions. Could it be that a different version of sphinx is used to generate the online docs, or that the typecheck hints were added after the current master?

jdcpni:
I don’t have any useful thoughts on these issue... not sure I fully understand them.  Happy to defer to both of your better judgement on the matters, though eager to better understand.  Unfortunately will be traveling during next meeting, but might be able to Skype Monday or Tuesday evening EDT if that would be useful.

jdcpni:
See comment added to PR.

jdc

Sent from my iPhone

On Jun 28, 2019, at 1:30 PM, Jan Vesely <notifications@github.com<mailto:notifications@github.com>> wrote:


@jdcpni<https://github.com/jdcpni>, @kmantel<https://github.com/kmantel>, this creates a lot of noise. I've checked that the output of
sphinx-build -b html -aE docs/source is almost the same with the only exception of removing type hints. Type hints produced output like:
<em class="sig-param">condition: &lt;typecheck.framework.optional object at 0x7f730808cba8&gt; = None</em>
but also:
<em class="sig-param">noise: psyneulink.core.globals.utilities.parameter_spec = 0.0</em>

It looks like typecheck could use some pretty-printing, so I disabled it ftm.
It might be fixed in later version (pnl is restricted to typecheck==1.2)
I've also some work in progress to use raw strings and avoid the need for extra escaping (backslashes), but I wanted to get this in early to avoid further conflicts.

jvesely:
found few examples of typecheck objects in docs:
https://princetonuniversity.github.io/PsyNeuLink/CombinationFunctions.html#psyneulink.core.components.functions.combinationfunctions.PredictionErrorDeltaFunction
https://princetonuniversity.github.io/PsyNeuLink/MemoryFunctions.html#psyneulink.core.components.functions.statefulfunctions.memoryfunctions.ContentAddressableMemory.add_to_memory
https://princetonuniversity.github.io/PsyNeuLink/ObjectiveFunctions.html#psyneulink.core.components.functions.objectivefunctions.Stability
https://princetonuniversity.github.io/PsyNeuLink/ObjectiveFunctions.html#psyneulink.core.components.functions.objectivefunctions.Stability

jvesely:
I can split the PR and submit the quote style separately.

I think that `is_preference_set` can be made to print something nicer than object address, typecheck would be more tricky and require monkey patching the class print methods.

kmantel:
I see, I bet there is something dependent on the exact version or environment the docs are generated from. Locally, the PredictionErrorMechanism example produces `prefs: psyneulink.core.globals.preferences.componentpreferenceset.is_pref_set` for me, while on the docs auto generated from master and evel it's ` prefs: <function is_pref_set at 0x1149c00d0> = None`. I think if this just disables those printing that makes sense to do. If we want to explain typehints we can just do it in the documentation for a given argument, or if we eventually move to just 3.7+ maybe the integrated typehinting will handle this cleaner

PR #1243, opened 2019-07-02
Reinitialize method update/Changed name of keyword in constructor of ModulatorySignals and its subclasses from "projections" to "modulates"

dillontsmith:
*Function/Mechanism Reinitialize Methods
-Changed the "reinitialize" methods of stateful Functions and Mechanisms with stateful Functions to apply to the component's last used execution id if no "execution_context" argument is specified. Previously, reinitialize methods would always apply to the None execution id in this case.
-Functions now set a most_recent_execution_id attribute and value parameter when they are called. This is to enable reinitialization and value retrieval at the Function level. 

*Modulatory Signals
-Changed name of keyword argument from "projections" to "modulates" for ModulatorySignal components and its subclasses.
-For the purposes of backwards compatibility and internal consistency with non-ModulatorySignal derived components, added "projections" as an unlisted alias for "modulates". If a user passes the old-style "projections" argument into a ModulatorySignal or ModulatorySignal-derived component, it is treated the same as if the "modulates" keyword was used.

*Scripts
-Updated existing scripts to use the now-standard "modulates" keyword argument instead of the old "projections" when constructing ModulatorySignal objects.

*Tests
-Updated existing tests to use the now-standard "modulates" keyword argument instead of the old "projections" when constructing ModulatorySignal objects.
-Wrote new tests to ensure equivalence of using "projections" vs "modulates" when constructing ModulatorySignal objects.


buildbot-princeton:

buildbot-princeton:

PR #1262, opened 2019-07-30
Travis docstrings avy

jvesely:
The list of error conditions:
http://www.pydocstyle.org/en/3.0.0/error_codes.html


jvesely:
> I'm ok with the style checking

Thanks!

PR #1268, opened 2019-08-02
feature: OCM modulation of nested node parameters

dillontsmith:
Created new Mechanism: ParameterInterfaceMechanism. It is a subclass of CompositionInterfaceMechanism that is granted special permissions to use ControlSignals as its OutputStates.

Compositions now instantiate with a ParameterInterfaceMechanism attribute.

Implemented feature allowing OCM of a composition to modulate parameter states of nested nodes via ParameterInterfaceMechanism. Necessary Projections, InputStates, and ControlSignals are added to compositions in _create_CIM_states().

Added _get_nested_nodes() and _get_nested_compositions() convenience methods to Composition.


buildbot-princeton:

PR #1276, opened 2019-08-23
Bug Fixes for PCIM Functionality

dillontsmith:
function:
-Corrected non-stateful use of modulation parameter

-Corrected non-stateful use of initialization_status parameter

composition:
-Previously, nested compositions executing within simulations would rely on their input projections to provide values to the input states of their input CIMs. Under the circumstance that a nested composition itself had an OCM and was executing within its own simulation, the projected value was invalid because the afferent node had not executed in that execution id. To fix this issue, Compositions executing within simulations now have their input values assigned to them under all circumstances.

-When validating input to be assigned to input_CIMs, uses default value instead of current value to find shape that must be matched by input. Using current value leads to crashes in cases where the input CIM has not yet executed.

tests:
-Added test for two-tier OCM modulation

-Updated test for LCControlMechanism to use Composition instead of System

### DIFF
--- a/psyneulink/core/components/functions/function.py
+++ b/psyneulink/core/components/functions/function.py
@@ -317,18 +317,16 @@ def _get_modulated_param(owner, mod_proj, execution_context=None):
     """
 
     from psyneulink.core.components.projections.modulatory.modulatoryprojection import ModulatoryProjection_Base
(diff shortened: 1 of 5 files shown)


buildbot-princeton:

buildbot-princeton:

PR #1294, opened 2019-08-30
Add codestyle checking to pytest

jvesely:


jvesely:
Most of the checks are disabled. I've enabled two that were easy to fix.
The process is rather straightforward:
1.) pick an error code and run
`~/Psyneulink $: pycodestyle --check=$ERROR_CODE`
2.) fix changes
3.) enable the error code check in `setup.cfg`
4.) commit

PR #1295, opened 2019-08-30
Bug fix - AutodiffComposition num_trials default value

dillontsmith:
Changed default value of num_trials from 1 to None for AutodiffComposition. Previous default caused autodiff comps to run only once in any situation where num_trials was not explicitly provided.


buildbot-princeton:

PR #1332, opened 2019-09-28
Codestyle: whitespace around operators

jvesely:
E227, E228. E222.
E226 is pretty invasive so it's not included in this PR.


jdcpni:
Seems fine to me

PR #1345, opened 2019-10-11
Model fit elfi

davidt0x:


davidt0x:
Looks like this failed on Travis because of time limit.

PR #1376, opened 2019-10-24
utilities: Simplify deepcopy with shared

jvesely:


jvesely:
Found during the investigation of pypy problems

PR #1384, opened 2019-11-08
Devel

dillontsmith:


jvesely:
> * immediately do a fast-forward merge of master into devel ( `git checkout devel && git merge --ff-only master`)

why was this part necessary? as long as the merge creates an empty merge commit, git should not see any conflicts in the next PR.

PR #1405, opened 2019-11-25
Refactor tutorial pytest

jvesely:
List notebook name as parameter.
Resue the same python interpreter as the one running the pytest.
Set `PYHTONPATH` env var to access the tested instance of psyneulink.


jvesely:
@kmantel, @jdcpni This refactors the tutorial test a bit which should make it more consistent with the rest of the pytest suite.

PR #1428, opened 2020-01-07
Avoid using Python 'random' module

jvesely:
Use numpy random instead.
Cleanup private random_state parameters.

### DIFF
--- a/tests/json/test_json.py
+++ b/tests/json/test_json.py
@@ -2,7 +2,6 @@ import numpy as np
 import os
 import psyneulink as pnl
 import pytest
-import random
 
 # stroop stimuli
 red = [1, 0]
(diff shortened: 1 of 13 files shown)


tests/json/test_json.py
```diff
@@ -48,7 +47,6 @@ def test_json_results_equivalence(
 
     # reset random seed
     seed = 0
-    random.seed(seed)
     np.random.seed(seed)
     pnl.core.globals.utilities.set_global_seed(seed)
```
kmantel: This line should stay, the other resets are OK to remove

kmantel: Sorry, thought the comment pointed to the line. Line 53/51 (`pnl.core.globals.utilities.set_global_seed(seed)`) should stay.

PR #1446, opened 2020-01-22
test fix for test_integrator_mechanism

loopylangur:
Hi,

The test `TestIntegratorNoise::test_integrator_accumulator_noise_fn` in `test_integrator_mechanism.py` fails intermittently when the seed is not set. I observed that the developers had to update the expected value in the assertion with a new value over several occasions in the past even with the seed set. 

I ran the tests about 60 times with different seeds and observed the range of values for `val`. It looks something like this:
image

The mean of `val` is about ` 0.1648065410516709` and it ranges from `-2.4416497795147944` to `3.0591995142489194`.

I think a better way to set this assertion might be to account for this range of values using a broader check in the assertion. This would avoid the trouble of changing the expected values in the future.

Please let me know if this looks reasonable or if you have other suggestions on how to deal with this scenario.

### DIFF
--- a/tests/mechanisms/test_integrator_mechanism.py
+++ b/tests/mechanisms/test_integrator_mechanism.py
@@ -1128,7 +1128,7 @@ class TestIntegratorNoise:
 
         val = float(I.execute(10))
 
-        np.testing.assert_allclose(val, -0.977277879876411)
+        np.testing.assert_allclose(val, 0.1648065410516709, atol=3.0, rtol=0.0)
 
     @pytest.mark.mechanism
     @pytest.mark.integrator_mechanism


buildbot-princeton:

buildbot-princeton:

kmantel:
Hi, thanks for your interest. During testing, we always use a fixed seed to ensure that we get the same values every time when using randomness. I'm actually surprised that this test passes with any frequency without a fixed seed. Outside of a fixed-seed environment, we shouldn't expect these tests to pass.

I think using such a large tolerance is not the way to go, because we would not be able to discern the noise from any underlying computation.

In general, we will sometimes update the expected value if we make a change that we know is good and has the side effect of changing the noise. In this case, the failing test is preferred.

loopylangur:
The test actually failed in all the 60 instances without the fixed seed. 

Do you have an intuition about how much deviation in `val` is expected for this test?

kmantel:
Not concretely, no, maybe @jdcpni does. But using defaults, `AccumulatorIntegrator` has an initial `previous_value` of 0 and with one execution will add a single sample from the `NormalDist` noise function. `NormalDist` by default uses a `mean` of 0 and `standard_deviation` of 1, so [this](https://www4c.wolframalpha.com/Calculate/MSP/MSP310812g8b708334aei93000061i516acd14a98i9?MSPStoreType=image/gif&s=13) from Wolfram seems reasonable to me

loopylangur:
I see. Could this then also indicate a possible bug?

kmantel:
It doesn't look like one to me - what do you think is incorrect?

jdcpni:
I agree with Katherine.  I don’t I’ve access to the repo at the moment, but my sense  is that the likelihood of getting the exact value tested in the assertion with a variance of 1 is highly unlikely and this not surprising over 60 samples.  As K indicated, the tests are not intended to validate the noise distribution, for which we rely on Numpy and assume is being properly generated, but rather to validate all the rest of the computations carried out.  For that reason, we want to control for the noise in the tests, which is why we use fixed seeds.

jdc

Sent from my iPhone

On Jan 22, 2020, at 8:13 PM, Katherine Mantel <notifications@github.com> wrote:

﻿

It doesn't look like one to me - what do you think is incorrect?

loopylangur:
Thanks for the response. I understand the intention. Would something like this make sense here?:

```python
def test_integrator_accumulator_noise_fn(self):
        noise=NormalDist()
        I = IntegratorMechanism(
            name='IntegratorMechanism',
            function=AccumulatorIntegrator(
                noise=noise
            ),
        )

        val = float(I.execute(10))

        np.testing.assert_allclose(val, noise)
```

This could be a more robust test and you need not to worry about the seeds at all. This way you wont need to update the tests in future as well. Let me know what you think.

jdcpni:
Perhaps I am not following/understanding, but the final line below seems to assert that the result of executing the IntegratorMechanism should be the same as the noise distribution function, which does not make sense.

jdc

On Jan 23, 2020, at 2:59 PM, loopylangur <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for the response. I understand the intention. Would something like this make sense here?:

def test_integrator_accumulator_noise_fn(self):
        noise=NormalDist()
        I = IntegratorMechanism(
            name='IntegratorMechanism',
            function=AccumulatorIntegrator(
                noise=noise
            ),
        )

        val = float(I.execute(10))

        np.testing.assert_allclose(val, noise)

This could be a more robust test and you need not to worry about the seeds at all. This way you wont need to update the tests in future as well. Let me know what you think.

loopylangur:
> Perhaps I am not following/understanding, but the final line below seems to assert that the result of executing the IntegratorMechanism should be the same as the noise distribution function, which does not make sense.

Just to clarify, i used ` noise=NormalDist()()`, not ` noise=NormalDist()`
So, noise is just a sample from the normal distribution and the last line is asserting that the result should be same as the noise

jvesely:
> Just to clarify, i used ` noise=NormalDist()()`, not ` noise=NormalDist()`
> So, noise is just a sample from the normal distribution and the last line is asserting that the result should be same as the noise

Such modification would change the nature of this test. The test is intended to run with function/callable as a noise parameter instead of a constant.

you can argue that using:
```
def noise_f():
    return 5
...
       noise=noise_f
...
```
is sufficient to test this, but that doesn't make much of a difference. See below.

Individual tests don't need to worry about seeds (unless it's a very special test), the testing environment is set up so that PRNGs generate sequences. This is pretty much standard in testing functionality that includes PRNGs.

in this specific case:
* if the functionality of `IntegratorMechanism` changes the test ***should*** fail.
* if the functionality of `AccumulatorIntegrator` changes the test ***should*** fail.
* if the functionality of `NormalDist` changes the test ***should*** fail.
* if the random sequence of numpy PRNG changes the test ***should*** fail.

some of the changes might be ok, or even desirable. In those cases, the tested value is updated and the original change is documented and explained in the modifying commit.
This behaviour is intentional to help developers see impact of changes they make.

loopylangur:
Thanks for the explanation. I think that makes sense for this scenario.

jvesely:
Thank you for your interest in pnl.
I've created few suggestions that might be a good introduction to contribution to PNL, I'm sure others will add more.

https://github.com/PrincetonUniversity/PsyNeuLink/issues?q=is%3Aopen+is%3Aissue+label%3Afirst-contribution

PR #1447, opened 2020-01-23
Fix compiled printf

SamKG:
@jvesely 
This pull fixes compiled printf by reverting it to the original version, and also adds extra test cases to better check for correctness.

There is a bug with the live version in which varargs are not correctly passed to the builtin printf wrapper


buildbot-princeton:

buildbot-princeton:

jvesely:
ok to test

jvesely:
add to whitelist

PR #1488, opened 2020-02-23
Refactor/autodiff

SamKG:
This refactor adds:

- composition.learn
    - Standard psyneulink learning and autodiffcomposition learning has been merged into one standardized api
- composition.disable_learning (instead of composition.enable_learning and composition.learning_enabled)
- A bugfix for hebbian learning 
- minibatching/patience/run_after(before)_minibatch functionality to both autodiff and psyneulink learning
- A partial restructure of the llvm compiled autodiff code
- cleanup of Autodiffcomposition code - Autodiff now uses the standard target_node input format, as used by normal learning in psyneulink
- helper methods that emulate the previous functionality of autodiff (i.e. automatic creation of backprop pathways)

### DIFF
--- a/psyneulink/core/compositions/composition.py
+++ b/psyneulink/core/compositions/composition.py
@@ -1875,9 +1875,9 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
         self._update_parameter_components()
 
         self.initialization_status = ContextFlags.INITIALIZED
-        #FIXME: This removes `composition.parameters.values`, as it was not being 
-        # populated correctly in the first place. `composition.parameters.results` 
-        # should be used instead - in the long run, we should look into possibly 
+        #FIXME: This removes `composition.parameters.values`, as it was not being
+        # populated correctly in the first place. `composition.parameters.results`
+        # should be used instead - in the long run, we should look into possibly
         # populating both values and results, as it would be more consistent with
         # the behavior of components
         del self.parameters.value
--- a/psyneulink/core/compositions/composition.py
+++ b/psyneulink/core/compositions/composition.py
@@ -6437,7 +6437,6 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
             animate=False,
             context=None,
             base_context=Context(execution_id=None),
-            learning_mode=False,
             ):
         """Pass inputs to Composition, then execute sets of nodes that are eligible to run until termination
         conditions are met.  See `Run` for details of formatting input specifications. See `Run` for details of
@@ -6847,7 +6846,6 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
                                         runtime_params=runtime_params,
                                         skip_initialization=True,
                                         bin_execute=bin_execute,
-                                        learning_mode=learning_mode
                                         )
 
             # ---------------------------------------------------------------------------------
@@ -6969,7 +6967,6 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
             clamp_input=SOFT_CLAMP,
             runtime_params=None,
             skip_initialization=False,
-            learning_mode=False,
             bin_execute=False,
             ):
         """
--- a/psyneulink/core/compositions/composition.py
+++ b/psyneulink/core/compositions/composition.py
@@ -795,24 +795,24 @@ Learning in a Composition
 
 Learning is used to modify the `Projections <Projection>` between Mechanisms in a Composition.  More specifically,
 it modifies the `matrix <MappingProjection.matrix>` parameter of those `MappingProjections <MappingProjection>`,
-which implements the strengths ("weights") of the associations between representations in the Mechanisms they connect. 
+which implements the strengths ("weights") of the associations between representations in the Mechanisms they connect.
 
 .. _Composition_Learning_Mode:
 
 *Running a Composition in Learning Mode*
 ======================================
-A Composition only learns when ran in learning mode, and when its `disable_learning` parameter is False. To run the Composition in learning mode, use the `learn <Composition.learn>` method. 
+A Composition only learns when ran in learning mode, and when its `disable_learning` parameter is False. To run the Composition in learning mode, use the `learn <Composition.learn>` method.
 See `learn <Composition.learn>` for more details.
 
 *Implementing Learning in a Composition*
 ======================================
-There are three ways of implementing learning in a Composition: 
+There are three ways of implementing learning in a Composition:
 
 i) using `standard PsyNeuLink Components <Composition_Learning_Standard>`
 
 ii) using the `AutodiffComposition <Composition_Learning_AutodiffComposition>` -- a specialized subclass of Composition that executes learning using `PyTorch <https://pytorch.org>`_
 
-iii) using `UserDefinedFunctions <UserDefinedFunction>`.  
+iii) using `UserDefinedFunctions <UserDefinedFunction>`.
 
 The advantage of using standard PsyNeuLink compoments is that it
 assigns each operation involved in learning to a dedicated Component. This helps make clear exactly what those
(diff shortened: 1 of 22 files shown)


psyneulink/core/llvm/execution.py
```diff
@@ -519,14 +515,13 @@ def _get_run_input_struct(self, inputs, num_input_sets):
...
         run_inputs = ((([iv] for m in origins for iv in inp[m][i]) for i in range(num_input_sets)) for inp in inputs)
-
         return c_input(*_tupleize(run_inputs))
 
     @property
     def _bin_run_func(self):
         if self.__bin_run_func is None:
             self.__bin_run_func = pnlvm.LLVMBinaryFunction.from_obj(
-                self._composition, tags=frozenset({"run"}))
+                self._composition, tags=frozenset({"run", *self.__additional_tags}))
```
jvesely: please use set manipulation routines.

psyneulink/library/compositions/compiledoptimizer.py
```diff
@@ -116,143 +117,143 @@ def step(self, ctx):
...
 
-                    delta_w_val = b2.load(b2.gep(delta_w_ptr,[zero, weight_row, weight_column]))
-                    pnlvm.helpers.printf(b2,"%f ",delta_w_val,override_debug=False)
-                pnlvm.helpers.printf(b1,"\n",override_debug=False)
+        #             delta_w_val = b2.load(b2.gep(delta_w_ptr,[zero, weight_row, weight_column]))
+        #             pnlvm.helpers.printf(b2,"%f ",delta_w_val,override_debug=False)
+        #         pnlvm.helpers.printf(b1,"\n",override_debug=False)
                 
-        pnlvm.helpers.printf(builder, f"\t\t\tOPTIM DONE UPDATE\n",override_debug=False)
+        # pnlvm.helpers.printf(builder, f"\t\t\tOPTIM DONE UPDATE\n",override_debug=False)
```
jvesely: This is a lot of dead code. Do we need to keep it around?

psyneulink/core/compositions/composition.py
```diff
@@ -6597,13 +6643,17 @@ def run(
             assert not is_simulation
             try:
                 if bin_execute is True or bin_execute.startswith('LLVM'):
-                    _comp_ex = pnlvm.CompExecution(self, [context.execution_id])
+                    comp_ex_tags = []
+                    # Add learning tags if the composition is to be ran in learning mode
+                    from psyneulink.library.compositions.autodiffcomposition import AutodiffComposition
+                    if self._is_learning(context) and isinstance(self, AutodiffComposition):
```
jvesely: Why does this need to check for autodiff instance? `isinstance` calls are rather expensive.
Use set for tags instead of list.
This looks like it will break CUDA execution.

psyneulink/library/compositions/compiledoptimizer.py
SamKG: this was commented out in error - it should not be so (will revert)

psyneulink/core/compositions/composition.py
SamKG: I wasn't sure if adding `learning` to tags would face an error in other compiled run situations, so I made it only add it in the case of autodiff

Can you elaborate on how it would break CUDA, and how I could possibly fix that? I can't test CUDA on my machine, so it would be good to know.

jvesely: There shouldn't be any compiled learning tests other than autodiff, so ti should be safe to drop. CUDA should match what LLVM. In this case adding the new tag in __ptx_initialize should be enough. you can ping me to test CUDA execution.

psyneulink/core/llvm/builder_context.py
```diff
@@ -307,39 +307,23 @@ def gen_autodiffcomp_learning_exec(self, composition, *, tags:frozenset):
...
-            data_out.attributes.remove('nonnull')
-
+        with self._gen_composition_exec_context(composition, tags=tags) as (builder, data, params, cond_gen):
+            state, _, comp_in, data, cond, = builder.function.args
             pytorch_model._gen_llvm_training_function_body(self, builder, state,
-                                                           params, data, learning)
+                                                           params, data)
             node_tags = tags.union({"node_wrapper"})
-            # Call output CIM
+            # # Call output CIM
```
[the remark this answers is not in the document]
SamKG: Typo - Fixed!

psyneulink/core/llvm/execution.py
```diff
@@ -567,6 +562,16 @@ def _initialize_autodiff_struct(self, autodiff_stimuli, arg_index,
...
+        if isgenerator(inputs):
+            list_inputs = list(inputs)
+            inputs = {k:[list(v)] for (k,v) in list_inputs[0].items()}
+            for i, inp in enumerate(list_inputs):
+                if i == 0:
+                    continue
+                for k,v in inp.items():
+                    inputs[k].append(v)
+            num_input_sets = len(next(iter(inputs.values())))
+            runs = num_input_sets
```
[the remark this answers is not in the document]
SamKG: Is there another way to handle generators without converting them to lists for compiled execution?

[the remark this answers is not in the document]
jvesely: I think it might be better to leave generators unsupported and let people use `LLVMExec`. The main advantage of generators is that not all values are in memory, and that is eliminated by creating a list.
If you want to support generators it'd be better to create a ctype structure directly.
The added code converts generator (list) of input sets into a dictionary of inputs for each input ndoe, which is then converted back to a list of single input per input node in `_get_run_input_struct`.

The second part was wrt `num_input_sets` isn't  the value the same as `len(list_inputs)`?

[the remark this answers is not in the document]
SamKG: ok I will look into replacing the generator code with a compiled equivalent of `compositionrunner` - for now, consider it a hack (will mark it as such). The main reason I went for `LLVMRun` instead of `LLVMExec` here was to get the additional gains from loop code-motion optimizations (as it previously did by running all epochs in a loop). I should test to see if they make a huge performance difference though.

I wasn't exactly sure what `num_input_sets` referred to - is it the same as the number of trials in the input?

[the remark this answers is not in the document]
jvesely: It'd be good to support both `LLVMExec` and `LLVMRun`. `LLVMRun` usually does perform better, but it depends on how many trials you want to run and how costly it is to set up. Going back and forth between nodes first vs. inputs first representation might take more time than `LLVMExec`.

`num_input_sets` is just a number of inputs. You can run more trials than `num_input_sets` in which case the run will loop over and start reusing inputs from the start. I'm not sure how it's supposed to work with generators, I'd expect that in this case `num_trials` and `num_input_sets` need to match.

It's not clear to me what `compositionrunner` should do. At a high level it looks like it's doing the same/similar job as a composition controller.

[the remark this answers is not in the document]
SamKG: `compositionrunner` is respnsible for batching inputs for learning cases. It also handles things like patience and epoch count.
The reasoning behind keeping it in a separate class was to isolate the compiled code, and also keep the methods specific to it from polluting `composition.py` (which is already quite a massive file)

psyneulink/core/llvm/execution.py
```diff
@@ -268,7 +269,8 @@ def _bin_func_multirun(self):
     def _set_bin_node(self, node):
         assert node in self._composition._all_nodes
         wrapper = self._composition._get_node_wrapper(node)
-        self.__bin_func = pnlvm.LLVMBinaryFunction.from_obj(wrapper, tags=frozenset({"node_wrapper"}))
+        tags = frozenset({"node_wrapper"}.union(self.__additional_tags))
```
jvesely: the other way around please: `self.__additional_tags.union(("node_wrapper"))`

psyneulink/core/compositions/composition.py
```diff
@@ -6598,14 +6642,17 @@ def run(
             # Simulations are run as part of the controller node wrapper.
             assert not is_simulation
             try:
+                comp_ex_tags = frozenset({})
+                # Add learning tags if the composition is to be ran in learning mode
+                if self._is_learning(context):
+                    comp_ex_tags = comp_ex_tags.union(frozenset({"learning"}))
```
jvesely: no need to create multiple frozensets and unions:
`comp_ex_tags = frozenset(("learning")) if self._is_learning(context) else frozenset()`
or even
`comp_ex_tags = frozenset(("learning") if self._is_learning(context) else ())`

PR #1535, opened 2020-03-18
llvm: (Fix) Add mod. inputs to input_type_list if len is >= 1

SamKG:
Fixes an issue where if there was only one modulatory input to a RTM, compilation would fail.

### DIFF
--- a/tests/composition/test_control.py
+++ b/tests/composition/test_control.py
@@ -911,7 +911,7 @@ class TestControlMechanisms:
                                       pytest.param("LLVMExec", marks=pytest.mark.llvm),
                                       pytest.param("LLVMRun", marks=pytest.mark.llvm),
                                      ])
-    def test_recurrent_control(self, mode, benchmark):
+    def test_recurrent_control(self, mode):
         monitor = pnl.TransferMechanism(default_variable=[[0.0]],
                                     size=1,
                                     function=pnl.Linear(slope=1, intercept=0),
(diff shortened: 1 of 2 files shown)


jvesely:
This needs a test

tests/composition/test_control.py
```diff
@@ -904,6 +904,45 @@ def test_multilevel_control(self, mode, benchmark):
...
         benchmark(oComp.run, [5], bin_execute=mode)
 
+    @pytest.mark.control
+    @pytest.mark.composition
+    @pytest.mark.parametrize("mode", ["Python",
+                                      pytest.param("LLVM", marks=pytest.mark.llvm),
+                                      pytest.param("LLVMExec", marks=pytest.mark.llvm),
+                                      pytest.param("LLVMRun", marks=pytest.mark.llvm),
+                                     ])
+    def test_recurrent_control(self, mode, benchmark):
```
jvesely: I'm surprised pytest doesn't complain about the missing `benchmark` mark. please drop the param, you're not suing benchmark anyway

tests/composition/test_control.py
```diff
@@ -904,6 +904,45 @@ def test_multilevel_control(self, mode, benchmark):
...
+        comp = pnl.Composition()
+        comp.add_node(monitor)
+        comp.add_node(rtm)
+        comp.add_node(controller)
+        comp.run(inputs = {
+                    monitor: [[1], [5], [1], [5]],
+                    rtm: [[1,0], [1,0] ,[1,0], [1,0]]
+                },
+            bin_execute=mode
+        )
```
jvesely: This should check the result. it'll probably require changing the modulated parameter to target one of the functions.

PR #1544, opened 2020-03-23
fix/warning/string_comp: To prevent unnecessary warnings, check if pa…

dillontsmith:
…rameter is a string before comparing with keyword

### DIFF
--- a/tests/projections/test_projection_specifications.py
+++ b/tests/projections/test_projection_specifications.py
@@ -1,7 +1,6 @@
 import psyneulink as pnl
 import numpy as np
 import pytest
-import warnings
 
 import psyneulink.core.components.functions.distributionfunctions
 import psyneulink.core.components.functions.statefulfunctions.integratorfunctions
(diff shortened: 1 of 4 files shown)


tests/projections/test_projection_specifications.py
```diff
@@ -349,6 +350,30 @@ def test_formats_for_gating_specification_of_input_and_output_ports(self):
...
+        p0 = pnl.MappingProjection(
+            matrix=[[0, 0, 0, 0],
+                    [0, 0, 0, 0],
+                    [0, 0, 0, 0],
+                    [0, 0, 0, 0]]
+        )
+        m1 = pnl.TransferMechanism(
+            default_variable=[0, 0, 0, 0]
+        )
+        with warnings.catch_warnings():
```
jvesely: this can just use `@pytest.mark.filterwarnigns("error:MyWarning")` for the test, or `pytest.warns()` to catch specific warning on a specific line instead of opencoding custom solution.

dillontsmith: Will update. Thanks for the suggestion.

jvesely:
retest this please

PR #1560, opened 2020-04-07
llvm: Extend `Composition` state and param structs to inherit from `C…

SamKG:
…omponent`


jvesely:
I think composition will also need to overlaod `_get_param_ids` and `_get_state_ids`.

SamKG:
retest this please

PR #1575, opened 2020-04-19
travis: add ppc64le jobs

jvesely:


jvesely:
looks like this will need to wait for https://travis-ci.community/t/disk-quota-exceeded-on-ppc64le/8006/3

PR #1589, opened 2020-04-28
Fix results shape from learning

SamKG:
This pull fixes a typo in `composition.learn`, which caused too many results to be returned when learning with minibatch sizes greater than 1.

### DIFF
--- a/psyneulink/library/compositions/compositionrunner.py
+++ b/psyneulink/library/compositions/compositionrunner.py
@@ -192,6 +192,9 @@ class CompositionRunner():
             if minibatch_size == TRAINING_SET:
                 minibatch_size = num_trials
 
+            if minibatch_size > num_trials:
+                raise Exception("The minibatch size cannot be greater than the number of trials.")
+
             early_stopper = None
             if patience is not None and (bin_execute is False or bin_execute == 'Python'):
                 early_stopper = EarlyStopping(min_delta=min_delta, patience=patience)
--- a/psyneulink/library/compositions/compositionrunner.py
+++ b/psyneulink/library/compositions/compositionrunner.py
@@ -201,14 +201,11 @@ class CompositionRunner():
             self._composition.run(inputs=minibatched_input, skip_initialization=skip_initialization, context=context, skip_analyze_graph=True, bin_execute=bin_execute, *args, **kwargs)
             skip_initialization = True
 
-        # FIXME: compiled run values differ from pytorch run
-        results = self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:] # return results from last epoch
-        # if bin_execute is not False and bin_execute != 'Python':
-        #     results = [x for x in self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:]] # return results from last epoch
-        # else:
-        #     results = [x[0] for x in self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:]] # return results from last epoch
+        num_epoch_results = num_trials // minibatch_size # number of results expected from final epoch
+        results = self._composition.parameters.results.get(context)[-1 * num_epoch_results:] # return results from last epoch
 
         return results
+
 class EarlyStopping(object):
     def __init__(self, mode='min', min_delta=0, patience=10):
         self.mode = mode
--- a/tests/api/test_api.py
+++ b/tests/api/test_api.py
@@ -9,3 +9,59 @@ class TestCompositionMethods:
         c.add_node(A)
         result = c.run(inputs={A: [1]}, num_trials=2)
         assert result == c.output_values == [np.array([1])]
+
+
+    # test whether xor model created as autodiff composition learns properly
+    @pytest.mark.pytorch
+    @pytest.mark.parametrize("mode", ['Python',
+                                      pytest.param('LLVMRun', marks=pytest.mark.llvm),
+                                     ])
+    @pytest.mark.parametrize("minibatch_size", [
+        1,
+        2,
+        3,
+        4
+    ])
+    def test_learning_output_shape(self, mode, minibatch_size):
+        '''
+        Tests for correct output from composition.learn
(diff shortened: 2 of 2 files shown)


psyneulink/library/compositions/compositionrunner.py
```diff
@@ -201,14 +201,11 @@ def run_learning(self,
             self._composition.run(inputs=minibatched_input, skip_initialization=skip_initialization, context=context, skip_analyze_graph=True, bin_execute=bin_execute, *args, **kwargs)
             skip_initialization = True
 
-        # FIXME: compiled run values differ from pytorch run
-        results = self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:] # return results from last epoch
-        # if bin_execute is not False and bin_execute != 'Python':
-        #     results = [x for x in self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:]] # return results from last epoch
-        # else:
-        #     results = [x[0] for x in self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:]] # return results from last epoch
+        num_epoch_results = int(num_trials / minibatch_size) # number of results expected from final epoch
```
jvesely, line 204: should this round up or down? use `//` for floor division

SamKG: should be truncated; will change to `//`

tests/api/test_api.py
```diff
@@ -9,3 +9,75 @@ def test_get_output_values_prop(self):
...
+        2,
+        3,
+        4
+    ])
+    def test_learning_output_shape(self, mode, minibatch_size):
+        '''
+        Tests for correct output from composition.learn
+        Expected: All results from last epoch
+        '''
+        num_epochs=2
```
jvesely, line 29: this look unused

psyneulink/library/compositions/compositionrunner.py
```diff
@@ -201,14 +201,11 @@ def run_learning(self,
             self._composition.run(inputs=minibatched_input, skip_initialization=skip_initialization, context=context, skip_analyze_graph=True, bin_execute=bin_execute, *args, **kwargs)
             skip_initialization = True
 
-        # FIXME: compiled run values differ from pytorch run
-        results = self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:] # return results from last epoch
-        # if bin_execute is not False and bin_execute != 'Python':
-        #     results = [x for x in self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:]] # return results from last epoch
-        # else:
-        #     results = [x[0] for x in self._composition.parameters.results.get(context)[-1 * num_trials * minibatch_size:]] # return results from last epoch
+        num_epoch_results = num_trials // minibatch_size # number of results expected from final epoch
```
jvesely, line 204: is `minibatch_size` guaranteed to be < `num_trials`? or is `num_epoch_results == 0` a legal outcome here?

tests/api/test_api.py
```diff
@@ -9,3 +9,75 @@ def test_get_output_values_prop(self):
...
+                                                                            hidden_comp,
+                                                                            hidden_to_out_comp,
+                                                                            output_comp],
+                                                                            learning_rate=10)
+        target_mech = backprop_pathway.target
+
+        results = xor_comp.learn(inputs={input_comp:xor_inputs,
+                                          target_mech:xor_targets},
+                                  epochs=10,
+                                  minibatch_size=minibatch_size
```
jvesely, line 80: this looks like it's missing `bin_execute=mode`, or is there other way to select compiled execution?

tests/api/test_api.py
```diff
@@ -9,3 +9,75 @@ def test_get_output_values_prop(self):
         c.add_node(A)
         result = c.run(inputs={A: [1]}, num_trials=2)
         assert result == c.output_values == [np.array([1])]
+
+
+    # test whether xor model created as autodiff composition learns properly
+    @pytest.mark.parametrize("mode", ['Python',
+                                      pytest.param('LLVMRun', marks=pytest.mark.llvm),
```
jvesely, line 16: What about `LLVMExec`? does it work? should it be tested as well?

psyneulink/library/compositions/compositionrunner.py
SamKG: `minibatch_size` should always be `<=` `num_trials`.

tests/api/test_api.py
SamKG: Ah, I changed the test case for this and in the process I moved it from using AutodiffCompositions -> Standard PNL Compositions.

Since standard PNL doesn't have compiled learning yet, I will have to convert it back to Autodiff - however, the test seemed to fail on the Windows run of GitHub Actions, as `AutodiffComposition` wasn't being successfully imported.

Is there a known fix for this issue?

tests/api/test_api.py
SamKG: Learning should only work using `LLVMRun` at present - it could theoretically be possible to support `LLVMExec` if a writeback step was added after compiled execute.

psyneulink/library/compositions/compositionrunner.py
jvesely: ok, I presume there's a check somewhere earlier enforcing it.

tests/api/test_api.py
jvesely: do you mean the 32 bit windows run?
if so, the test should be tagged `@pytest.mark.pytorch` so ti can be skipped if pytorch is not available.

tests/api/test_api.py
jvesely: sounds good. `LLVMExec` support can be added later for easier debugging, but I don't expect anyone to need it.

psyneulink/library/compositions/compositionrunner.py
SamKG: added a check

Issue #1545, opened 2020-03-23
intermittent failures in `test_xor_training_correctness[LLVMRun-50-single-adam-False]`


jvesely:
see for example:
https://ci.appveyor.com/project/jvesely/psyneulink-coveralls/builds/31663125
 a repeated build of the same code succeeds:
https://ci.appveyor.com/project/jvesely/psyneulink-coveralls/builds/31663917

jvesely:
cc @SamKG

SamKG:
Will take a look at this - I suspect this bug was introduced when I had removed ref passing for the weights

jvesely:
I checked the test, and it doesn't hit the path fixed in #1610.
Also, the results look wrong even for Python mode, printing them shows:
```
tests/composition/test_autodiffcomposition.py::TestTrainingCorrectness::test_xor_training_correctness[Python-50-single-adam-True] [[array([0.31200036])], [array([0.59406178])], [array([0.60417587])], [array([0.52347365])]]
PASSED
tests/composition/test_autodiffcomposition.py::TestTrainingCorrectness::test_xor_training_correctness[Python-50-single-adam-False] [[array([0.399898])], [array([0.58961467])], [array([0.54293491])], [array([0.51298241])]]
PASSED
tests/composition/test_autodiffcomposition.py::TestTrainingCorrectness::test_xor_training_correctness[LLVMRun-50-single-adam-True] [[[0.31202413946754987]], [[0.5940729063910729]], [[0.6041634977020353]], [[0.5234295358656476]]]
PASSED
tests/composition/test_autodiffcomposition.py::TestTrainingCorrectness::test_xor_training_correctness[LLVMRun-50-single-adam-False] [[[0.43985781659105583]], [[0.5190865881628404]], [[0.5582109612930329]], [[0.5144072619250126]]]
PASSED
```

while the `True` results look similar between Python and LLVM. The `False` results differ widely.

Not that the 4th result rounds to `1` even though it's expected to be `0`.
The tests pass because the results comparison loop checks only the first result and ignores the rest.

SamKG:
I'll take another pass on verifying correctness; The gap between Python and LLVM looks way larger than previously seen, so it's possible there is some discrepancy between the logic in both.

kmantel:
I ran into this on the 100 variant, see 
[...]

PR #1610, opened 2020-05-09
autodiffcomposition: Don't use python random

jvesely:


SamKG:
Is the seed for normal python random never set?

This could be the cause of the elusive xor correctness issue in the autodiff (#1545)

jvesely:
> Is the seed for standard library random not set to a fixed value for tests?

No, it was removed with other uses of python random in #1428
> 
> This could be the cause of the elusive xor correctness issue (#1545)

Does the test use randomized epochs?

PR #1613, opened 2020-05-10
tests/xor_training_correctness: Test all results against target

jvesely:
Increase number of training epochs to fix '1 xor 1' result.
Codestyle.


jvesely:
potential fix for #1545

PR #1621, opened 2020-05-15
Refactor/composition/preprocessing

dillontsmith:
- Ports system functionality relating to input and output labels to Composition
- _adjust_stimulus_dict, _adjust_execution_stimuli, and other input parsing methods that were called in the run method have been removed and replaced by a series of more concise parsing methods 
- Adds ability to specify input dict as function in call to Composition.learn
- Passing something falsy as the `inputs` argument to Composition.run or Composition.execute is now a legal call, and will result in the Composition using its nodes' default variables


tests/composition/test_composition.py
```diff
@@ -2641,6 +2641,38 @@ def test_LPP_wrong_component(self):
                 and "to InputPort" in str(error_text.value)
                 and "that is in deferred init" in str(error_text.value))
 
+    @pytest.mark.composition
+    @pytest.mark.parametrize("mode", ['Python',
+                                      pytest.param('LLVM', marks=pytest.mark.llvm),
+                                      pytest.param('LLVMExec', marks=pytest.mark.llvm),
+                                      pytest.param('LLVMRun', marks=pytest.mark.llvm),
+                                      pytest.param('PTXExec', marks=[pytest.mark.llvm, pytest.mark.cuda]),
+                                      pytest.param('PTXRun', marks=[pytest.mark.llvm, pytest.mark.cuda])
```
jvesely, line 2650: Please drop `*Run` from the list. It doesn't do anything for `execute`

psyneulink/core/compositions/composition.py
```diff
@@ -8769,38 +9201,52 @@ def execute(
...
+            call_before_pass : callable : default None
                 called before each `PASS` is executed
                 passed the current *context* (but it is not necessary for your callable to take)
 
-            call_after_pass : callable
+            call_after_pass : callable : default None
                 called after each `PASS` is executed
                 passed the current *context* (but it is not necessary for your callable to take)
 
+            bin_execute : bool or Enum[LLVM|LLVMexec|LLVMRun|Python|PTXExec|PTXRun] : default Python
```
jvesely, line 9246: I think the `*Run` options should be dropped from here as well. Don't do anything when passed to `execute`.

tests/composition/test_composition.py
dillontsmith: Updated

psyneulink/core/compositions/composition.py
dillontsmith: Updated

dillontsmith:
retest this please

PR #1624, opened 2020-05-19
tests/composition,learning: Extend stimuli length mismatch test

jvesely:
Allow any order of reported input lengths.


jvesely:
the original version fails on PyPy which always preserves insertion order even for sets.

PR #1632, opened 2020-05-21
setup: Update required/supported python versions

jvesely:
Bump coverage requirement to 75%


kmantel:
This is good. We use f-strings a lot now, which were only added in 3.6. I'm not sure if there is anything else.

Increasing coverage requirements as we meet them also seems good.

PR #1633, opened 2020-05-21
documentation/composition/stateful_reset

dillontsmith:
- documentation/composition/stateful_reset
    - Updated documentation for specifying reset values in Composition.run
- fix/composition/stateful_reset:
    - Fixed bug causing reset_stateful_functions_to values not to propagate from Composition.run to Composition.execute


dillontsmith:
@jvesely there is a compilation-related method in `Composition` called `reset`, that conflicts with another `reset` method in Composition. The compilation-related method doesn't seem to be used anywhere, and commenting it out doesn't break any checks. Is this method needed? If so, can it be renamed?

jvesely:
> @jvesely there is a compilation-related method in `Composition` called `reset`, that conflicts with another method named `reset` in `Composition`. The compilation-related method doesn't seem to be used anywhere, and commenting it out doesn't break any checks. Is this method needed? If so, can it be renamed?

It's safe to drop. It was used to reset binary structures associated with a specific `execution_id`, but the test that needed it was converted to use unique execution ids instead.

psyneulink/core/compositions/composition.py
```diff
@@ -9631,7 +9796,12 @@ def execute(
                             context=context
                         )
                     ):
-                        node.reset(None, context=context)
+                        if node in reset_stateful_functions_to:
+                            node.reset(*reset_stateful_functions_to[node],
+                                       context=context)
+                        else:
+                            node.reset(None,
+                                       context=context)
```
jvesely, line 9804:
```
vals = reset_state_functions_to.get(node, [None])
node.reset(*vals, context=context)
```
would save one lookup in the successful case

psyneulink/core/compositions/composition.py
```diff
@@ -10057,12 +10227,16 @@ def _update_learning_parameters(self, context):
...
+        if context is NotImplemented:
+            context = self.most_recent_context
 
-        for i in range(self.stateful_nodes):
-            self.stateful_nodes[i].reset(values[i], context=context)
+        if not values:
+            values = {}
+
+        for node in self.stateful_nodes:
+            reset_val = None if not node in values else values[node]
```
jvesely, line 10238: same here (`None` is the default reutnred by `get`)
`reset_val = values.get(node)`

psyneulink/core/compositions/composition.py
```diff
@@ -10325,16 +10499,16 @@ def _gen_llvm_function(self, *, ctx:pnlvm.LLVMBuilderContext, tags:frozenset):
...
+    # @handle_external_context(execution_id=NotImplemented)
+    # def reset(self, context=None):
+    #     if context.execution_id is NotImplemented:
+    #         context.execution_id = self.most_recent_context.execution_id
+    #
+    #     self._compilation_data.ptx_execution.set(None, context)
+    #     self._compilation_data.parameter_struct.set(None, context)
+    #     self._compilation_data.state_struct.set(None, context)
+    #     self._compilation_data.data_struct.set(None, context)
+    #     self._compilation_data.scheduler_conditions.set(None, context)
```
jvesely, line 10511: yeah, this can be dropped. the test was migrated long ago. Not that this is the only way to cleanup binary structures associated with a specific execution id.

psyneulink/core/compositions/composition.py
```diff
@@ -9037,16 +9204,19 @@ def run(
...
-
-        for node in reset_stateful_functions_to:
-            # make sure the args to reset are in the form of an iterable so they can be unpacked
-            try:
-                iter(reset_stateful_functions_to[node])
-            except TypeError:
-                reset_stateful_functions_to[node] = [reset_stateful_functions_to[node]]
-            node.reset(*reset_stateful_functions_to[node], context=context)
+        try:
+            for node in reset_stateful_functions_to:
```
jvesely, line 9208:
```
for node, vals in reset_state_functions_to.items():
     ...
```
can avoid several repeated lookups

psyneulink/core/compositions/composition.py
```diff
@@ -9037,16 +9204,19 @@ def run(
...
+                # make sure the args to reset are in the form of an iterable so they can be unpacked
+                try:
+                    iter(reset_stateful_functions_to[node])
+                except TypeError:
+                    reset_stateful_functions_to[node] = [reset_stateful_functions_to[node]]
```
jvesely, line 9213: the conversion can be put in the if-block below to avoid unnecessary conversion if it's not used

psyneulink/core/compositions/composition.py
```diff
@@ -9037,16 +9204,19 @@ def run(
...
+                if (isinstance(reset_stateful_functions_when, Never) or
+                        node not in reset_stateful_functions_when) and \
+                        isinstance(node.reset_stateful_function_when, Never):
+                    node.reset(*reset_stateful_functions_to[node], context=context)
+        except TypeError:
```
jvesely, line 9218: should the entire loop exit if one iterations raises TypeError?

psyneulink/core/compositions/composition.py
dillontsmith: updated in ffad3efb0747fedd8306bad83a2bfc5dd5ddee29

psyneulink/core/compositions/composition.py
dillontsmith: updated in ffad3efb0747fedd8306bad83a2bfc5dd5ddee29

psyneulink/core/compositions/composition.py
dillontsmith: updated in ffad3efb0747fedd8306bad83a2bfc5dd5ddee29

psyneulink/core/compositions/composition.py
dillontsmith: updated in ffad3efb0747fedd8306bad83a2bfc5dd5ddee29

psyneulink/core/compositions/composition.py
dillontsmith: In this case, the conversion shouldn't be in the if-block, because the converted values will be used later, regardless of whether or not conditions are met to trigger the reset here.

psyneulink/core/compositions/composition.py
dillontsmith: This was previously being used to catch `TypeError`s arising from when `reset_stateful_functions_to` was `None`, but the block was restructured in ffad3efb0747fedd8306bad83a2bfc5dd5ddee29 and no longer does this.

dillontsmith:
retest this please

PR #1646, opened 2020-05-29
Context: Change SIMULATION from an execution_phase to a runmode

kmantel:
- SIMULATION is now renamed to SIMULATION_MODE
- Context.execution_phase can now only have one flag
- Context.runmode can now have two flags only if SIMULATION is one of them
- calls to `add_flag` for Context.execution_phase are replaced with simple attribute setting

### DIFF
--- a/psyneulink/core/compositions/composition.py
+++ b/psyneulink/core/compositions/composition.py
@@ -9659,7 +9659,7 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
                 self._reset_stateful_functions_when_cache[node] = node.reset_stateful_function_when
                 node.reset_stateful_function_when = reset_stateful_functions_when[node]
 
-        if ContextFlags.SIMULATION not in context.execution_phase:
+        if ContextFlags.SIMULATION not in context.runmode:
             try:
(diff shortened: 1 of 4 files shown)


jdcpni:
Not sure if K replied, but the answer is No.  The idea is that the execution_phase flags are ones that are sensibly mutually exclusive (IDLE, PREPARING, PROCESSING) (should probably rename PROCESSING to EXECUTING).  The modes are conditions that apply to those states, individually or in combination.  So, yes, LEARNING_MODE could co-occur with SIMULATION_MODE if a simulation is run using learning.  There is a subtler issue, that remains to be resolved, of how to handle the LEARNING execution phase.  Properly, this should be a mode, but it requires finer handling than LEARNING_MODE.  So, we may need to create additional, finer-grained modes, or perhaps another category.  This is something we should discuss and implement soon.


jdc



On May 29, 2020, at 3:40 PM, SamKG <notifications@github.com<mailto:notifications@github.com>> wrote:



@SamKG commented on this pull request.

Looks good!
Would this change mean that simulation mode and learning mode are now mutually exclusive?
Is there ever a situation in which we would want both?

PR #1655, opened 2020-06-02
Composition/Controller/HangingControlSpecs

dillontsmith:
Nodes with specifications for control that were specified at the Mechanism level can now connect to controllers of enclosing Compositions in cases where their immediate Composition does not have a controller.


dillontsmith:
retest this please

PR #1657, opened 2020-06-09
Fix learning correctness

SamKG:
- Fix correctness issues in learning (now should be numerically exact)
- Reorder operations in the ADAM optimizer to conform to the exact ordering used in PyTorch
- Cleanup compiled learning code + setup for compiled RNN
- Add additional tests for autodiff correctness
- Remove autodiff.get_parameters (obsoleted by the ability to get matrices directly from the projections)

### DIFF
--- a/psyneulink/core/components/functions/transferfunctions.py
+++ b/psyneulink/core/components/functions/transferfunctions.py
@@ -731,8 +731,12 @@ class Exponential(TransferFunction):  # ----------------------------------------
 
 
         """
-        return self._get_current_function_param(RATE, context) * input + self._get_current_function_param(BIAS, context)
+        from math import e
+        rate = self._get_current_function_param(RATE, context)
+        scale = self._get_current_function_param(SCALE, context)
+        bias = self._get_current_function_param(BIAS, context)
 
+        return rate * scale * e**(rate * input + bias)
 
 # **********************************************************************************************************************
 #                                                   Logistic
--- a/psyneulink/core/components/functions/transferfunctions.py
+++ b/psyneulink/core/components/functions/transferfunctions.py
@@ -1241,7 +1241,6 @@ class Tanh(TransferFunction):  # -----------------------------------------------
         bias_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, BIAS)
         x_0_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, X_0)
         offset_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, OFFSET)
-        scale_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, SCALE)
 
         gain = pnlvm.helpers.load_extract_scalar_array_one(builder, gain_ptr)
         bias = pnlvm.helpers.load_extract_scalar_array_one(builder, bias_ptr)
--- a/psyneulink/core/components/functions/transferfunctions.py
+++ b/psyneulink/core/components/functions/transferfunctions.py
@@ -1511,7 +1511,11 @@ class ReLU(TransferFunction):  # -----------------------------------------------
         gain = self._get_current_function_param(GAIN, context)
         leak = self._get_current_function_param(LEAK, context)
 
-        return gain if input > 0 else gain * leak
+        input = np.asarray(input).copy()
+        input[input>0] = gain
+        input[input<=0] = gain * leak
+
+        return input
 
 
 # **********************************************************************************************************************
--- a/psyneulink/core/components/functions/transferfunctions.py
+++ b/psyneulink/core/components/functions/transferfunctions.py
@@ -68,7 +68,7 @@ from psyneulink.core.globals.keywords import \
 from psyneulink.core.globals.parameters import \
     Parameter, get_validator_by_function
 from psyneulink.core.globals.utilities import parameter_spec, get_global_seed, safe_len
-from psyneulink.core.globals.context import ContextFlags
+from psyneulink.core.globals.context import ContextFlags, handle_external_context
 from psyneulink.core.globals.preferences.basepreferenceset import \
     REPORT_OUTPUT_PREF, PreferenceEntry, PreferenceLevel, is_pref_set
 
@@ -122,10 +122,10 @@ class TransferFunction(Function_Base):
                 vo = b.gep(arg_out, [ctx.int32_ty(0), idx])
                 with pnlvm.helpers.array_ptr_loop(b, vi, "nested_transfer_loop") as args:
                     self._gen_llvm_transfer(ctx=ctx, vi=vi, vo=vo,
-                                            params=params, state=state, *args)
+                                            params=params, state=state, *args, tags=tags)
             else:
                self._gen_llvm_transfer(b, idx, ctx=ctx, vi=arg_in, vo=arg_out,
-                                       params=params, state=state)
+                                       params=params, state=state, tags=tags)
 
         return builder
 
@@ -384,7 +384,7 @@ class Linear(TransferFunction):  # ---------------------------------------------
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
         slope_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, SLOPE)
@@ -393,9 +393,15 @@ class Linear(TransferFunction):  # ---------------------------------------------
         slope = pnlvm.helpers.load_extract_scalar_array_one(builder, slope_ptr)
         intercept = pnlvm.helpers.load_extract_scalar_array_one(builder, intercept_ptr)
 
-        val = builder.load(ptri)
-        val = builder.fmul(val, slope)
-        val = builder.fadd(val, intercept)
+
+        if "derivative" in tags:
+            # f'(x) = m
+            val = slope
+        else:
+            # f(x) = mx + b
+            val = builder.load(ptri)
+            val = builder.fmul(val, slope)
+            val = builder.fadd(val, intercept)
 
         builder.store(val, ptro)
 
@@ -451,6 +457,7 @@ class Linear(TransferFunction):  # ---------------------------------------------
 
         return self.convert_output_type(result)
 
+    @handle_external_context()
     def derivative(self, input=None, output=None, context=None):
         """
         derivative(input)
@@ -638,7 +645,7 @@ class Exponential(TransferFunction):  # ----------------------------------------
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
@@ -657,8 +664,15 @@ class Exponential(TransferFunction):  # ----------------------------------------
         val = builder.fmul(val, rate)
         val = builder.fadd(val, bias)
         val = builder.call(exp_f, [val])
-        val = builder.fmul(val, scale)
-        val = builder.fadd(val, offset)
+
+        if "derivative" in tags:
+            # f'(x) = s*r*e^(r*x + b)
+            val = builder.fmul(val, scale)
+            val = builder.fmul(val, rate)
+        else:
+            # f(x) = s*e^(r*x + b) + o
+            val = builder.fmul(val, scale)
+            val = builder.fadd(val, offset)
 
         builder.store(val, ptro)
 
@@ -697,6 +711,7 @@ class Exponential(TransferFunction):  # ----------------------------------------
         result = scale * e**(rate * variable + bias) + offset
         return self.convert_output_type(result)
 
+    @handle_external_context()
     def derivative(self, input, output=None, context=None):
         """
         derivative(input)
@@ -908,7 +923,7 @@ class Logistic(TransferFunction):  # -------------------------------------------
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
@@ -926,6 +941,7 @@ class Logistic(TransferFunction):  # -------------------------------------------
 
         exp_f = ctx.get_builtin("exp", [ctx.float_ty])
         val = builder.load(ptri)
+
         val = builder.fadd(val, bias)
         val = builder.fsub(val, x_0)
         val = builder.fmul(val, gain)
@@ -934,6 +950,14 @@ class Logistic(TransferFunction):  # -------------------------------------------
         val = builder.fadd(ctx.float_ty(1), val)
         val = builder.fdiv(ctx.float_ty(1), val)
         val = builder.fmul(val, scale)
+
+        if "derivative" in tags:
+            # f(x) = g * s * o * (1-o)
+            function_val = val
+            val = builder.fsub(ctx.float_ty(1), function_val)
+            val = builder.fmul(function_val, val)
+            val = builder.fmul(gain, val)
+            val = builder.fmul(scale, val)
 
         builder.store(val, ptro)
 
@@ -974,6 +998,7 @@ class Logistic(TransferFunction):  # -------------------------------------------
 
         return self.convert_output_type(result)
 
+    @handle_external_context()
     def derivative(self, input=None, output=None, context=None):
         """
         derivative(input=None, output=None)
@@ -1204,7 +1229,7 @@ class Tanh(TransferFunction):  # -----------------------------------------------
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
@@ -1212,12 +1237,16 @@ class Tanh(TransferFunction):  # -----------------------------------------------
         bias_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, BIAS)
         x_0_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, X_0)
         offset_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, OFFSET)
+        scale_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, SCALE)
 
         gain = pnlvm.helpers.load_extract_scalar_array_one(builder, gain_ptr)
         bias = pnlvm.helpers.load_extract_scalar_array_one(builder, bias_ptr)
         x_0 = pnlvm.helpers.load_extract_scalar_array_one(builder, x_0_ptr)
         offset = pnlvm.helpers.load_extract_scalar_array_one(builder, offset_ptr)
         exp_f = ctx.get_builtin("exp", [ctx.float_ty])
+
+        assert "derivative" not in tags, f"Compiled derivatives are not currently supported for {self}!"
+
         exp_val = builder.load(ptri)
         exp_val = builder.fadd(exp_val, bias)
         exp_val = builder.fsub(exp_val, x_0)
@@ -1271,6 +1300,7 @@ class Tanh(TransferFunction):  # -----------------------------------------------
         return self.convert_output_type(result)
 
 
+    @handle_external_context()
     def derivative(self, input, output=None, context=None):
         """
         derivative(input)
@@ -1467,7 +1497,7 @@ class ReLU(TransferFunction):  # -----------------------------------------------
 
         return self.convert_output_type(result)
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
@@ -1482,14 +1512,20 @@ class ReLU(TransferFunction):  # -----------------------------------------------
         # Maxnum for some reason needs full function prototype
         max_f = ctx.get_builtin("maxnum", [ctx.float_ty])
         var = builder.load(ptri)
-        val = builder.fsub(var, bias)
-        val1 = builder.fmul(val, gain)
-        val2 = builder.fmul(val1, leak)
 
-        val = builder.call(max_f, [val1, val2])
+        if "derivative" in tags:
+            predicate = builder.fcmp_ordered('>', var, var.type(0))
+            val = builder.select(predicate, gain, builder.fmul(gain, leak))
+        else:
+            val = builder.fsub(var, bias)
+            val1 = builder.fmul(val, gain)
+            val2 = builder.fmul(val1, leak)
+
+            val = builder.call(max_f, [val1, val2])
 
         builder.store(val, ptro)
 
+    @handle_external_context()
     def derivative(self, input, output=None, context=None):
         """
         derivative(input)
@@ -1678,7 +1714,7 @@ class Gaussian(TransferFunction):  # -------------------------------------------
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
@@ -1751,6 +1787,7 @@ class Gaussian(TransferFunction):  # -------------------------------------------
 
         return self.convert_output_type(result)
 
+    @handle_external_context()
     def derivative(self, input, output=None, context=None):
         """
         derivative(input)
@@ -1961,7 +1998,7 @@ class GaussianDistort(TransferFunction):  #-------------------------------------
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags:frozenset):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
@@ -2425,6 +2462,7 @@ class SoftMax(TransferFunction):
 
         return self.convert_output_type(output)
 
+    @handle_external_context()
     def derivative(self, output, input=None, context=None):
         """
         derivative(output)
--- a/psyneulink/core/components/mechanisms/mechanism.py
+++ b/psyneulink/core/components/mechanisms/mechanism.py
@@ -2945,8 +2945,8 @@ class Mechanism_Base(Mechanism):
                                        mech_params, mech_state, mech_in)
         return builder
 
-    def _gen_llvm_invoke_function(self, ctx, builder, function, params, state, variable):
-        fun = ctx.import_llvm_function(function)
+    def _gen_llvm_invoke_function(self, ctx, builder, function, params, state, variable, *, tags:frozenset):
+        fun = ctx.import_llvm_function(function, tags=tags)
         fun_in, builder = self._gen_llvm_function_input_parse(builder, ctx, fun, variable)
         fun_out = builder.alloca(fun.args[3].type.pointee)
 
@@ -2957,7 +2957,7 @@ class Mechanism_Base(Mechanism):
     def _gen_llvm_is_finished_cond(self, ctx, builder, params, state, current):
         return pnlvm.ir.IntType(1)(1)
 
-    def _gen_llvm_function_internal(self, ctx, builder, params, state, arg_in, arg_out):
+    def _gen_llvm_function_internal(self, ctx, builder, params, state, arg_in, arg_out, *, tags:frozenset):
 
         ip_output, builder = self._gen_llvm_input_ports(ctx, builder,
                                                         params, state, arg_in)
--- a/psyneulink/core/components/mechanisms/mechanism.py
+++ b/psyneulink/core/components/mechanisms/mechanism.py
@@ -2967,7 +2967,7 @@ class Mechanism_Base(Mechanism):
                 self.function, f_params_ptr, ctx, builder, params, state, arg_in)
 
         f_state = pnlvm.helpers.get_state_ptr(builder, self, state, "function")
-        value, builder = self._gen_llvm_invoke_function(ctx, builder, self.function, f_params, f_state, ip_output)
+        value, builder = self._gen_llvm_invoke_function(ctx, builder, self.function, f_params, f_state, ip_output, tags=tags)
 
         # Update execution counter
         exec_count_ptr = pnlvm.helpers.get_state_ptr(builder, self, state, "execution_count")
@@ -3030,7 +3030,7 @@ class Mechanism_Base(Mechanism):
                                                     return_type=pnlvm.ir.IntType(1))
         iparams, istate, iin, iout = internal_builder.function.args[:4]
         internal_builder, is_finished = self._gen_llvm_function_internal(ctx, internal_builder,
-                                                                         iparams, istate, iin, iout)
+                                                                         iparams, istate, iin, iout, tags=tags)
         internal_builder.ret(is_finished)
 
         # Call Internal Function
--- a/psyneulink/core/compositions/composition.py
+++ b/psyneulink/core/compositions/composition.py
@@ -6624,7 +6624,9 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
                         p.insert(0, curr_node)
                         curr_node = prev[curr_node]
                     p.insert(0, curr_node)
-                    pathways.append(p)
+                    # we only consider input -> projection -> ... -> output pathways (since we can't learn on only one mechanism)
+                    if len(p) >= 3:
+                        pathways.append(p)
                     continue
                 for projection, efferent_node in [(p, p.receiver.owner) for p in curr_node.efferents]:
                     if (not hasattr(projection,'learnable')) or (projection.learnable is False):
--- a/psyneulink/core/llvm/execution.py
+++ b/psyneulink/core/llvm/execution.py
@@ -151,9 +151,9 @@ class CUDAExecution:
 
 class FuncExecution(CUDAExecution):
 
-    def __init__(self, component, execution_ids=[None]):
+    def __init__(self, component, execution_ids=[None], *, tags=frozenset()):
         super().__init__()
-        self._bin_func = pnlvm.LLVMBinaryFunction.from_obj(component)
+        self._bin_func = pnlvm.LLVMBinaryFunction.from_obj(component, tags=tags)
         self._execution_contexts = [
             Context(execution_id=eid) for eid in execution_ids
         ]
--- a/psyneulink/library/compositions/autodiffcomposition.py
+++ b/psyneulink/library/compositions/autodiffcomposition.py
@@ -143,7 +143,6 @@ from psyneulink.core import llvm as pnlvm
 import copy
 import numpy as np
 import ctypes
-import warnings
 from collections.abc import Iterable
 from toposort import toposort
 from inspect import isgenerator
@@ -281,17 +280,11 @@ class AutodiffComposition(Composition):
     def _build_pytorch_representation(self, context=None):
         if self.scheduler is None:
             self.scheduler = Scheduler(graph=self.graph_processing)
-        if self.execution_sets is None:
-            self.execution_sets = [ x - set(self.get_nodes_by_role(NodeRole.LEARNING)) for x in self.scheduler.run(context=context)]
-            self.execution_sets = [x for x in self.execution_sets if len(x) > 0]
-        if self.parameters.pytorch_representation._get(context) is None:
-            model = PytorchModelCreator(self.graph_processing,
-                                        self.param_init_from_pnl,
-                                        self.execution_sets,
-                                        self.device,
-                                        context=context,
-                                        composition = self,
-                                        )
+        if self.parameters.pytorch_representation._get(context=context) is None:
+            model = PytorchModelCreator(composition=self,
+                                        device=self.device,
+                                        context=context)
+
             self.parameters.pytorch_representation._set(model, context, skip_history=True, skip_log=True)
 
         # Set up optimizer function
@@ -379,7 +372,6 @@ class AutodiffComposition(Composition):
         curr_tensor_outputs = self.parameters.pytorch_representation._get(context).forward(
             curr_tensor_inputs,
             context,
-            scheduler=scheduler,
         )
 
         for component in curr_tensor_outputs.keys():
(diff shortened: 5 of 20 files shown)


psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+    def add_afferent(self, afferent):
+        if not afferent in self.afferents:
+            self.afferents.append(afferent)
+
+
+    def collate_afferents(self):
+        """
+        Returns weight-multiplied sum of all afferent projections
+        """
+        return sum([proj.execute(proj.sender.value) for proj in self.afferents])
```
jvesely, line 150: a generator would be better than list comprehension.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def execute(self, variable):
+        self.value = self.function(variable)
+        # print(f"{self} input: {variable}\n")
+        # print(f"{self} output: {self.value}\n")
+        return self.value
+
+    def execute_llvm(self, ctx, builder, state, params, mech_input, data):
```
jvesely, line 158: This method is not executing things. It's generating the execution call.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+        builder.call(mech_func, [mech_param, mech_state,
+                                    mech_input, mech_output])
+        
+        for i in range(len(mech_input.type.pointee.elements[0])):
+            pnlvm.helpers.printf_float_array(builder, builder.gep(mech_input, [ctx.int32_ty(0), ctx.int32_ty(0), ctx.int32_ty(i)]), prefix=f"{self} input:\n", override_debug=False)
+        pnlvm.helpers.printf_float_array(builder, builder.gep(mech_output, [ctx.int32_ty(0), ctx.int32_ty(0)]), prefix=f"{self} output:\n", override_debug=False)
+
+        return mech_output
+
+    def execute_derivative_func_llvm(self, ctx, builder, mech_input):
```
jvesely, line 182: This is not executing functions. It's just generating the call.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+    def get_afferent_idx(self, afferent_node):
+        return [proj.receiver for proj in self.afferents].index(afferent_node)
+
+    def get_efferent_idx(self, efferent_node):
+        return [proj.sender for proj in self.efferents].index(efferent_node)
+
+    def __repr__(self): 
+        return "PytorchWrapper for: "+self._mechanism.__repr__() 
+
+class PytorchProjection():
```
jvesely, line 196: It's confusing to name things Projection if they don't inherit from `Projection`

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+        return torch.matmul(variable, self.matrix)
+
+    def _extract_llvm_matrix(self, ctx, builder, params):
+        proj_params = builder.gep(params, [ctx.int32_ty(0),
+                                           ctx.int32_ty(1),
+                                           ctx.int32_ty(self._idx)])
+
+        dim_x, dim_y = self.matrix.detach().numpy().shape
+        proj_matrix = pnlvm.helpers.get_param_ptr(builder, self._projection, proj_params, "matrix")
+        proj_matrix = builder.bitcast(proj_matrix, pnlvm.ir.types.ArrayType(
```
jvesely, line 236: pls no `bitcast`. It breaks and invalidates a lot of things and should not be used anywhere.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+            pnlvm.ir.types.ArrayType(ctx.float_ty, dim_y), dim_x).as_pointer())
+
+        return proj_matrix
+
+    def execute_llvm(self, ctx, builder, state, params, data):
```
jvesely, line 241: This method is not executing anything. it just generates the call.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+        mech_state = builder.gep(state, [ctx.int32_ty(0),
+                                            ctx.int32_ty(0),
+                                            ctx.int32_ty(self._idx)])
+
+        mech_output = builder.gep(data, [ctx.int32_ty(0),
+                                            ctx.int32_ty(0),
+                                            ctx.int32_ty(self._idx)])
+
+        builder.call(mech_func, [mech_param, mech_state,
+                                    mech_input, mech_output])
```
jvesely, line 174: Inconsistent indentation.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+
+        matrix = projection.parameters.matrix.get(
+                            context=context)
+        if matrix is None:
+            matrix = projection.parameters.matrix.get(
+                context=None
+            )
+        self.matrix = torch.nn.Parameter(torch.tensor(matrix.copy(),
+                                   device=device,
+                                   dtype=torch.double))
```
jvesely, line 217: indentation looks wrong.

tests/composition/test_autodiffcomposition.py
```diff
@@ -62,6 +62,38 @@ def test_report_prefs(self):
         # comp = AutodiffComposition()
         # assert comp.patience == 10
 
+class TestNonTrainingFunctionality:
```
jvesely, line 65: Why does this need a class?

tests/composition/test_autodiffcomposition.py
```diff
@@ -62,6 +62,38 @@ def test_report_prefs(self):
...
+    @pytest.mark.parametrize("mode", ['Python',
+                                      pytest.param('LLVMRun', marks=pytest.mark.llvm),
+                                     ])
```
jvesely, line 68: this should be marked pytorch. otherwise, it will fail without the `torch` module

tests/composition/test_autodiffcomposition.py
```diff
@@ -466,7 +421,7 @@ class TestTrainingCorrectness:
         'eps, calls, opt, from_pnl_or_not, expected', [
             (100, 'single', 'adam', True, [[[0.09823965]], [[0.81092879]], [[0.78179557]], [[0.25593583]]]),
             (50, 'multiple', 'adam', True, [[[0.31200036]], [[0.59406178]], [[0.60417587]], [[0.52347365]]]),
-            (100, 'single', 'adam', False, [[[0.12697489]], [[0.74632817]], [[0.80712739]], [[0.28699516]]]),
+            (1, 'single', 'adam', True, [[[0.94790852]], [[0.95989353]], [[0.94048928]], [[0.95685814]]]),
```
jvesely, line 424: why do we need to test both `100-single-adam-True` and `1-single-adam-True`?
Why does this remove `single-adam-false`?
The `False` variant suffered more from lack of accuracy.

psyneulink/library/compositions/compiledoptimizer.py
```diff
@@ -207,51 +166,57 @@ def step(self, ctx):
             gen_inject_mat_add(ctx, builder, v_t_ptr, delta_w_sqrd, v_t_ptr)
 
         # 4) update weights
+        # this is the new learning rate to use
+        # NOTE: This differs from the version specified in the paper to numerically match with pytorch's implementation
+        step_size = builder.fdiv(lr, one_minus_b1_pow)
+        step_size = builder.fmul(step_size, ctx.float_ty(-1.0))
```
jvesely, line 172: `helpers.fneg`

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+        return lambda x: (torch.max(input=(x - bias), other=torch.tensor([0], device=device).double()) * gain +
+                            torch.min(input=(x - bias), other=torch.tensor([0], device=device).double()) * leak)
+
+    else:
+        raise Exception(f"Function {function} is not currently supported in AutodiffCompositions!")
+
+# Helper method that creates a bin func that returns the derivative of the function into the builder
+# FIXME: Add compiled derivative functions, and move these calls there
+def bin_function_derivative_creator(ctx, node, context=None):
+    # first try to get cached func
```
jvesely, line 48: The function should be generated from an object rather than a name, caching then works transparently and no of this is needed.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+try:
+    import torch
+    from torch import nn
+    torch_available = True
+except ImportError:
+    torch_available = False
+
+
+# helper method that identifies the type of function used by a node, gets the function
+# parameters and uses them to create a function object representing the function, then returns it
```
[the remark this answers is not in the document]
SamKG: it is required for the pytorch specific operators (e.g. torch.exp), which allows for pytorch to calculate the gradient from the computation graph.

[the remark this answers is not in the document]
jvesely: so pytorch can handle `pytorch.exp` but not `np.exp` of `**`?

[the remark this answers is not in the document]
SamKG: yes, I believe so. seems like it's an active effort (https://github.com/pytorch/pytorch/issues/2228)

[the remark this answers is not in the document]
jvesely: (pytorch/pytorch#2228) refers to introducing numpy like API in pytorch, it doesn't talk about supporting bumpy types and operations. what am I missing?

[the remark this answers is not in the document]
SamKG: see:https://github.com/pytorch/pytorch/issues/2228#issuecomment-488808650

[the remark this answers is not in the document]
jvesely: that talks about adding a compatible API rather than supporting operations from other packages.
using `torch.numpy` would add hard requirement on pytorch and break all configurations that are not supported by pytorch.

I'm ok with this transformation, just add a comment explaining that it's due to torch limitations.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,257 @@
...
+                                            ctx.int32_ty(self._idx)])
+
+        mech_output = builder.gep(data, [ctx.int32_ty(0),
+                                            ctx.int32_ty(0),
+                                            ctx.int32_ty(self._idx)])
+
+        builder.call(mech_func, [mech_param, mech_state,
+                                    mech_input, mech_output])
+        
+        for i in range(len(mech_input.type.pointee.elements[0])):
```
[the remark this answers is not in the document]
SamKG: `enumerate(x)` isn't possible in this case, since the type is of ir.types.ArrayType, which is not iterable. `len()` in this case actually just uses a magic method on the class, which is why it works.

I'll make it more explicit by moving the `len()` call out of the `range()` call.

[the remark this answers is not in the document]
jvesely: ir.ArrayType has `count` property, that should be more readable than `range(len(` (even if split)

psyneulink/library/compositions/pytorchcomponents.py
SamKG: this would require removing the dimensionality assertions in the `_gen_inject_<op>` helpers, as previously used, since matrix is a flattened parameter (and hence is of type double*, instead of an array of array of doubles)

not sure if that would be preferable

psyneulink/library/compositions/pytorchcomponents.py
SamKG: trying to implement this by adding a class for the derivative function, but I seem to be getting a segfault due to a parsing error; 
any pointers on how I should approach this?

jvesely: using `PNL_LLVM_DEBUG=llvm` will dump the llvm assembly files (~one per function) you can then check which function generates invalid syntax

tests/composition/test_autodiffcomposition.py
SamKG: Remove `param_init_pnl`, as it was deprecated. This should remove the issue with the False cases, as they were irrelevant anyways.

psyneulink/library/compositions/pytorchcomponents.py
jvesely: it might alos be better to just use `node` and select derivative using `tags`

tests/composition/test_autodiffcomposition.py
jvesely: so why does it keep `50-multiple-adam-False` ?

SamKG: whoops typo in previous comment -  meant to say "Remove**d**" instead of "Remove".
Changes are pushed - will squash fixups when everything is approved.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,276 @@
...
+from psyneulink.library.compositions.pytorchllvmhelper import *
+from psyneulink.core.globals.log import LogCondition
+from psyneulink.core import llvm as pnlvm
+
+try:
+    import torch
+    from torch import nn
+    torch_available = True
+except ImportError:
+    torch_available = False
```
jvesely, line 11: `torch_available` is not used anywhere in this file

psyneulink/library/compositions/pytorchcomponents.py
SamKG: will flag as future cleanup

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,276 @@
...
+def pytorch_function_creator(function, device, context=None):
+    """
+    Converts a PsyNeuLink function into an equivalent PyTorch lambda function.
+    NOTE: This is needed due to PyTorch limitations (see: https://github.com/PrincetonUniversity/PsyNeuLink/pull/1657#discussion_r437489990)
+    """
+    def get_fct_param_value(param_name):
+        val = function._get_current_function_param(
+            param_name, context=context)
+        if val is None:
+            val = getattr(function.defaults, param_name)
```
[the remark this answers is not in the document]
SamKG: It appears so - `_get_current_function_param` doesn't seem to return the default value (not sure if that is intended).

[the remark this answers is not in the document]
jvesely: ok, thanks for checking

tests/composition/test_autodiffcomposition.py
```diff
@@ -379,101 +407,23 @@ def test_params_stay_separate(self,mode):
...
     @pytest.mark.parametrize(
-        'eps, calls, opt, from_pnl_or_not, expected', [
-            (100, 'single', 'adam', True, [[[0.09823965]], [[0.81092879]], [[0.78179557]], [[0.25593583]]]),
-            (50, 'multiple', 'adam', True, [[[0.31200036]], [[0.59406178]], [[0.60417587]], [[0.52347365]]]),
-            (100, 'single', 'adam', False, [[[0.12697489]], [[0.74632817]], [[0.80712739]], [[0.28699516]]]),
-            (50, 'multiple', 'adam', False, [[[0.2935138]], [[0.60503794]], [[0.57901045]], [[0.57705371]]])
+        'eps, calls, opt, expected', [
+            (100, 'single', 'adam', [[[0.09823965]], [[0.81092879]], [[0.78179557]], [[0.25593583]]]),
+            (50, 'multiple', 'adam', [[[0.31200036]], [[0.59406178]], [[0.60417587]], [[0.52347365]]]),
+            (1, 'single', 'adam', [[[0.94790852]], [[0.95989353]], [[0.94048928]], [[0.95685814]]]),
```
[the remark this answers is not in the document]
SamKG: Useful as a specific case for debugging - will remove it though, since it doesn't serve a purpose generally.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,276 @@
...
+                                         ctx.int32_ty(self._idx)])
+
+        builder.call(mech_func, [mech_param,
+                                 mech_state,
+                                 mech_input,
+                                 mech_output])
+
+        #NOTE: range(len()) is not an antipattern here, since mech_input.type.pointee.elements[0] is not enumerable
+        input_length = mech_input.type.pointee.elements[0].count
+        for i in range(input_length):
```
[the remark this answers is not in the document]
SamKG: I'll just remove it altogether, as it's meant to only exist for debug purposes anyways.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def _gen_execute_llvm(self, ctx, builder, state, params, data):
+        proj_matrix = self._extract_llvm_matrix(ctx, builder, params)
+
+        input_vec = builder.gep(data, [ctx.int32_ty(0),
+                                       ctx.int32_ty(0),
+                                       ctx.int32_ty(self.sender._idx),
+                                       ctx.int32_ty(self._port_idx)])
+
+        output_vec = gen_inject_vxm(ctx, builder, input_vec, proj_matrix)
```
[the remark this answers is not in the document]
SamKG: this code should be uneeded when the forward computation is changed to reuse the existing composition execution - for now, I have left it as-is.

psyneulink/library/compositions/pytorchcomponents.py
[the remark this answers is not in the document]
jvesely: I checked parameter `_get` and it needs `parameter.fallback_default` to get that behaviour. is that not the case for these parameters? if not, should it be?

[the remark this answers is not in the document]
SamKG: It doesn't seem to be the case - `return getattr(self.parameters, param_name)._get(context)` seems to be returning `None` instead of the default value of the parameter (seems to be due to contexts). Seems like a bug.

SamKG:
retest this please

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,276 @@
...
+    #       2) sizeof vector
+    #       3) ptr to output vector
+    float_ptr_ty = ctx.float_ty.as_pointer()
+    args = [float_ptr_ty, ctx.int32_ty, float_ptr_ty]
+    builder = ctx.create_llvm_function(args, node, name)
+    llvm_func = builder.function
+
+    input_vector, dim, output_vector = llvm_func.args
+
+    def get_fct_param_value(param_name):
```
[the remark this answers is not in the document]
SamKG: Solved in latest commit

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -122,10 +122,10 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, arg_in, arg_out,
                 vo = b.gep(arg_out, [ctx.int32_ty(0), idx])
                 with pnlvm.helpers.array_ptr_loop(b, vi, "nested_transfer_loop") as args:
                     self._gen_llvm_transfer(ctx=ctx, vi=vi, vo=vo,
-                                            params=params, state=state, *args)
+                                                params=params, state=state, *args, tags=tags)
```
jvesely, line 125: spurious indentation change

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -384,7 +384,7 @@ def __init__(self,
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, tags:frozenset):
```
jvesely, line 387: `tags` should be compulsory keyword arg

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -638,7 +644,7 @@ def __init__(self,
...
```
jvesely, line 647: `tags` should be compulsory keyword arg

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -908,7 +925,7 @@ def __init__(self,
...
```
jvesely, line 928: `tags` should be compulsory keyword arg

psyneulink/core/components/functions/function.py
```diff
@@ -627,10 +627,10 @@ def _get_current_function_param(self, param_name, context=None):
                                 f"the current value of a function parameter. 'variable' is not a function parameter. "
                                 f"If looking for {self.name}'s default variable, try {self.name}.defaults.variable.")
         try:
-            return self.owner._parameter_ports[param_name].parameters.value._get(context)
+            return self.owner._parameter_ports[param_name].parameters.value.get(context)
         except (AttributeError, TypeError):
             try:
-                return getattr(self.parameters, param_name)._get(context)
+                return getattr(self.parameters, param_name).get(context)
```
jvesely, line 633: should this get @kmantel 's review?

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1204,31 +1230,37 @@ def __init__(self,
...
```
jvesely, line 1233: `tags` should be compulsory keyword arg

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1204,31 +1230,37 @@ def __init__(self,
...
         bias_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, BIAS)
         x_0_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, X_0)
         offset_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, OFFSET)
+        scale_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, SCALE)
 
         gain = pnlvm.helpers.load_extract_scalar_array_one(builder, gain_ptr)
         bias = pnlvm.helpers.load_extract_scalar_array_one(builder, bias_ptr)
         x_0 = pnlvm.helpers.load_extract_scalar_array_one(builder, x_0_ptr)
         offset = pnlvm.helpers.load_extract_scalar_array_one(builder, offset_ptr)
+        scale = pnlvm.helpers.load_extract_scalar_array_one(builder, scale_ptr)
```
jvesely, line 1247: `scale` is unused in the computation below

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1204,31 +1230,37 @@ def __init__(self,
...
-        exp_val = builder.fadd(exp_val, offset)
-        exp_val = builder.fmul(exp_val.type(-2), exp_val)
-
-        val = builder.call(exp_f, [exp_val])
-        val1 = builder.fsub(val.type(1), val)
-        val2 = builder.fadd(val.type(1), val)
-        val = builder.fdiv(val1, val2)
+
+        if "derivative" in tags:
+            raise Exception(f"Compiled derivatives are not currently supported for {self}!")
```
jvesely, line 1251: I don't think this needs to be a user-friendly error. just use `assert`

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1482,13 +1514,23 @@ def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
...
 
-        val = builder.call(max_f, [val1, val2])
+        if "derivative" in tags:
+            predicate = builder.fcmp_ordered('>', var, var.type(0))
+            with builder.if_else(predicate) as (then, otherwise):
+                with then:
+                    builder.store(gain, ptro)
+                with otherwise:
+                    val = builder.fmul(gain, leak)
+                    builder.store(val, ptro)
```
jvesely, line 1525: `select` [0] would be more readable here

[0] https://llvmlite.readthedocs.io/en/latest/user-guide/ir/ir-builder.html#llvmlite.ir.IRBuilder.select

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1467,7 +1499,7 @@ def _function(self,
 
         return self.convert_output_type(result)
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, tags:frozenset):
```
jvesely, line 1502: `tags` should be compulsory keyword arg

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1511,7 +1553,11 @@ def derivative(self, input, output=None, context=None):
         gain = self._get_current_function_param(GAIN, context)
         leak = self._get_current_function_param(LEAK, context)
 
-        return gain if input > 0 else gain * leak
+        input = np.asarray(input).copy()
```
jvesely, line 1556: why does this need a copy? you're dropping the reference to original in the same statement.

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1674,7 +1720,7 @@ def __init__(self,
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, tags:frozenset):
```
jvesely, line 1723: `tags` should be compulsory keyword arg

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1957,7 +2003,7 @@ def __init__(self,
...
```
jvesely, line 2006: `tags` should be compulsory keyword arg

psyneulink/core/components/mechanisms/mechanism.py
```diff
@@ -2944,8 +2944,8 @@ def _fill_input(b, s_input, i):
                                        mech_params, mech_state, mech_in)
         return builder
 
-    def _gen_llvm_invoke_function(self, ctx, builder, function, params, state, variable):
-        fun = ctx.import_llvm_function(function)
+    def _gen_llvm_invoke_function(self, ctx, builder, function, params, state, variable, tags = frozenset()):
```
jvesely, line 2947: `tags` should be compulsory keyword arg

psyneulink/core/compositions/composition.py
```diff
@@ -6612,7 +6612,8 @@ def bfs(start):
                         p.insert(0, curr_node)
                         curr_node = prev[curr_node]
                     p.insert(0, curr_node)
-                    pathways.append(p)
+                    if len(p) >= 3:
```
jvesely, line 6615: this could use a comment about the `3` magic number

psyneulink/core/llvm/execution.py
```diff
@@ -151,9 +151,9 @@ def cuda_execute(self, variable):
 
 class FuncExecution(CUDAExecution):
 
-    def __init__(self, component, execution_ids=[None]):
+    def __init__(self, component, execution_ids=[None], tags = frozenset()):
```
jvesely, line 154: `tags` should be compulsory keyword arg

arg assignment shouldn't be surrounded by whitespace

psyneulink/library/compositions/autodiffcomposition.py
```diff
@@ -281,16 +274,10 @@ def __init__(self,
...
-            self.execution_sets = [ x - set(self.get_nodes_by_role(NodeRole.LEARNING)) for x in self.scheduler.run(context=context)]
-            self.execution_sets = [x for x in self.execution_sets if len(x) > 0]
-        if self.parameters.pytorch_representation._get(context) is None:
-            model = PytorchModelCreator(self.graph_processing,
-                                        self.param_init_from_pnl,
-                                        self.execution_sets,
-                                        self.device,
+        if self.parameters.pytorch_representation._get(context=context) is None:
+            model = PytorchModelCreator(composition = self,
+                                        device = self.device,
```
jvesely, line 279: arg assignment should be surrounded by whitespace

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,209 @@
...
+        self.value = None
+        self._llvm_value = None
+        self.afferents = []
+        self.efferents = []
+
+        self._target_mechanism = None
+
+    def add_efferent(self, efferent):
+        if not efferent in self.efferents:
+            self.efferents.append(efferent)
```
jvesely, line 64: the interface is weird. is it legal to insert the same efferent twice?

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def add_afferent(self, afferent):
+        if not afferent in self.afferents:
+            self.afferents.append(afferent)
```
jvesely, line 68: the interface is weird. is it legal to insert the same afferent twice?

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,209 @@
...
+        Returns weight-multiplied sum of all afferent projections
+        """
+        return sum((proj.execute(proj.sender.value) for proj in self.afferents))
+
+    def execute(self, variable):
+        self.value = self.function(variable)
+
+        return self.value
+
+    def _gen_execute_llvm(self, ctx, builder, state, params, mech_input, data):
```
jvesely, line 82: the convention is to use `_gen_llvm_` prefix

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,209 @@
...
+
+        return mech_output
+
+    def log_value(self):
+        if self._mechanism.parameters.value.log_condition != LogCondition.OFF:
+            detached_value = self.value.detach().cpu().numpy()
+            self._mechanism.output_port.parameters.value._set(detached_value, self._context)
+            self._mechanism.parameters.value._set(detached_value, self._context)
+
+    def _gen_execute_derivative_func_llvm(self, ctx, builder, state, params, data, arg_in):
```
jvesely, line 112: the convention is to use `_gen_llvm_` prefix

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,209 @@
...
+        # HACK: pack input in bigger struct
```
jvesely, line 113: why is this hack needed? it looks like it tries to match whatever input shape to that of the mechanism function, but uses custom array instead of the actual input shape.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -0,0 +1,209 @@
...
+                                           ctx.int32_ty(self._idx)])
+
+        dim_x, dim_y = self.matrix.detach().numpy().shape
+        proj_matrix = pnlvm.helpers.get_param_ptr(builder, self._projection, proj_params, "matrix")
+        proj_matrix = builder.bitcast(proj_matrix, pnlvm.ir.types.ArrayType(
+            pnlvm.ir.types.ArrayType(ctx.float_ty, dim_y), dim_x).as_pointer())
+
+        return proj_matrix
+
+    def _gen_execute_llvm(self, ctx, builder, state, params, data):
```
jvesely, line 192: the convention is to use `_gen_llvm_` prefix

tests/functions/test_transfer.py
```diff
@@ -90,6 +99,23 @@ def test_execute(func, variable, params, fail, expected, benchmark, mode):
     assert np.allclose(res, expected)
     benchmark(f.function, variable)
 
+@pytest.mark.function
+@pytest.mark.transfer_function
+@pytest.mark.parametrize("func, variable, params, expected", derivative_test_data)
```
jvesely, line 104: pls add names/ids to derivative test data

tests/functions/test_transfer.py
```diff
@@ -90,6 +99,23 @@ def test_execute(func, variable, params, fail, expected, benchmark, mode):
...
+def test_execute_derivative(func, variable, params, expected, mode):
+    f = func(default_variable=variable, **params)
+    if mode == 'Python':
+        ex = f.derivative
+    elif mode == 'LLVM':
+        ex = pnlvm.execution.FuncExecution(f, tags=frozenset({"derivative"})).execute
+    elif mode == 'PTX':
+        ex = pnlvm.execution.FuncExecution(f, tags=frozenset({"derivative"})).cuda_execute
+    res = ex(variable)
+    assert np.allclose(res, expected)
```
jvesely, line 118: pls add benchmark after the assert check

psyneulink/core/components/functions/transferfunctions.py
SamKG: no copy is performed if the input was already a numpy array (see: https://numpy.org/doc/stable/reference/generated/numpy.asarray.html)

jvesely: so? what do you need to call `copy()`?

psyneulink/library/compositions/pytorchcomponents.py
SamKG: Shouldn't be legal (would be akin to adding the same projection twice)

psyneulink/core/llvm/execution.py
SamKG: this change would be inconsistent with the existing paradigm for `CompExecution`:

`class CompExecution(CUDAExecution):
    def __init__(self, composition, execution_ids=[None], additional_tags=frozenset()):`

not sure if we wanted to change that constructor as well

psyneulink/core/components/functions/function.py
SamKG: yes - will revert for now (ideally should be a separate issue)

psyneulink/core/components/functions/transferfunctions.py
SamKG: wanted to prevent inplace modification of the array (incase it messed with stored state in the broader context of PsyNeuLink)

if this isn't an issue, I can remove the copy

psyneulink/library/compositions/pytorchcomponents.py
SamKG: is there any way to construct the expected type of the function?
I tried digging around the `_parse_input_ports` method in Mechanism, but was confused as to what the actual structure should be.

psyneulink/core/llvm/execution.py
jvesely: I'd say that `additional_tags` should follow the same pattern as `tags`, but that can be done separately.

psyneulink/library/compositions/autodiffcomposition.py
jvesely: sorry that should should ahve been "should not be surrounded by whitespace". sorry for the confusion.
basically following https://www.flake8rules.com/rules/E251.html, in before it can be enforced globally

psyneulink/library/compositions/pytorchcomponents.py
jvesely: `set` doesn't have guaranteed iteration order so that might interfere with indices in compilation.
maybe an assert check would suffice.

psyneulink/library/compositions/pytorchcomponents.py
jvesely: you can import the function and check `args[0].type` if there needs to be a type conversion.
do you know why this happens? sounds like another example of shape mismatch. do python shapes match?

tests/functions/test_transfer.py
```diff
@@ -90,6 +106,25 @@ def test_execute(func, variable, params, fail, expected, benchmark, mode):
     assert np.allclose(res, expected)
     benchmark(f.function, variable)
 
+@pytest.mark.function
+@pytest.mark.transfer_function
+@pytest.mark.transfer_function
```
jvesely, line 111: transfer_function mark is added twice.

it'd be nice to mark as benchmark and assign benchmark group.

psyneulink/core/llvm/execution.py
SamKG: https://github.com/PrincetonUniversity/PsyNeuLink/issues/1695

made an issue for it

jvesely: the issue doesn't make much sense to me. `tags` argument should be added when/as needed.
maybe I phrased it wrong; 'compulsory keyword arg` == `tags` should be keywords-only argument when added.

SamKG: ah, i see what you meant. alright, will modify the issue to reflect this.

psyneulink/core/components/functions/transferfunctions.py
jvesely: ah, so the situation to avoid is the input is already an np.ndarray and the lines below would modify it. Yeah, that makes sense

psyneulink/core/llvm/execution.py
jvesely: thanks. it also applies to all other `tags` arguments. it should always be keyword-only.

tests/functions/test_transfer.py
```diff
@@ -90,6 +106,26 @@ def test_execute(func, variable, params, fail, expected, benchmark, mode):
...
+    benchmark.group = "TransferFunction " + func.componentName + " Derivative"
+    if mode == 'Python':
+        ex = f.derivative
+    elif mode == 'LLVM':
+        ex = pnlvm.execution.FuncExecution(f, tags=frozenset({"derivative"})).execute
+    elif mode == 'PTX':
+        ex = pnlvm.execution.FuncExecution(f, tags=frozenset({"derivative"})).cuda_execute
+    res = ex(variable)
+    assert np.allclose(res, expected)
+    benchmark(f.derivative, variable)
```
jvesely, line 128: This should benchmark `ex` instead of always running the python version

psyneulink/library/compositions/pytorchcomponents.py
SamKG: the function expects a 2d array as input, but I have a vector.

I suspect that this is due to input ports

SamKG:
retest this please

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1204,20 +1233,24 @@ def __init__(self,
...
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, tags):
         ptri = builder.gep(vi, [ctx.int32_ty(0), index])
         ptro = builder.gep(vo, [ctx.int32_ty(0), index])
 
         gain_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, GAIN)
         bias_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, BIAS)
         x_0_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, X_0)
         offset_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, OFFSET)
+        scale_ptr = pnlvm.helpers.get_param_ptr(builder, self, params, SCALE)
```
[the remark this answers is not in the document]
SamKG: Scale seems to only be used in the derivative (which is weird)

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -384,7 +384,7 @@ def __init__(self,
             prefs=prefs,
         )
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, tags):
```
jvesely, line 387: `tags` should be a keyword only argument

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1467,7 +1501,7 @@ def _function(self,
 
         return self.convert_output_type(result)
 
-    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state):
+    def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, tags):
```
jvesely, line 1504: `tags` should be a keyword only argument

psyneulink/core/components/functions/transferfunctions.py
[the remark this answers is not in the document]
jvesely: Let's drop it from here for now. The diff to fix the implementation will look nicer when it's added in both Python and LLVM, and we won't introduce a dead variable.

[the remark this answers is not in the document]
SamKG: sounds good - i have a patch in progress, should be done today.

Issue #1479, opened 2020-02-16
Function instance of `TransferMechanism.termination_measure` is shared among all instances.


jvesely:
the parameter is declared:
```
termination_measure = Parameter(Distance(metric=MAX_ABS_DIFF), modulable=False, stateful=False, loggable=False)
```
since it's a class attribute of the Prameters class the instance of `Distance` is shared among all instances of TransferMechanism.

kmantel:
In this case, the problem isn't that the `Distance` instance is used as a class default for `termination_measure`, it's that there is an instance created as the default argument to `termination_measure` in `__init__`: https://github.com/PrincetonUniversity/PsyNeuLink/blob/6c7fb550fbe2efb25abe59950864409ca1d4590c/psyneulink/core/components/mechanisms/processing/transfermechanism.py#L951

This is created once on import, and then used for every TransferMechanism without that argument specified. In general, the exact object passed in to a constructor will be used as the parameter value. The exception is for the `function` parameter, which would make a copy of the object if it was already used by another `Mechanism` (the function's `owner` is not None).

A quick fix would be to just set the default argument value in `TransferMechanism.__init__` above to None. This would have the default set properly for `termination_measure`, (And in general, they should be None, so the `_user_specified` Parameter flag is set correctly, so this would be a good change regardless) but it would still leave this example with the current behavior:

```
>>> dist = pnl.Distance()
>>> a = pnl.TransferMechanism(termination_measure=dist)
>>> b = pnl.TransferMechanism(termination_measure=dist)
>>> a.defaults.termination_measure is b.defaults.termination_measure
True
```

I think in general though, the `function` behavior should be expanded to other "subfunction" parameters

jvesely:
is the change in init a one-liner that should work, or does it trigger changes elsewhere? Do you have plans to work on it?
The example is OK. It probably won't work with compilation, but the sharing is clearly intended.

kmantel:
The change should work, because the Parameter default will be used when the argument to __init__ is None. Using None should be the correct pattern in general because it will let us know if the user passed in an argument or not. It's been a low priority task to go through and change all those, but it can be bumped up as necessary.

I guess I'm not sure what's the best thing to do with the example. For `function` we definitely wanted protection from instances being shared; I'm not sure if the same pattern should hold for other parameters that happen to take functions.

jvesely:
Fixes in ded82bf

PR #1666, opened 2020-06-12
Enforce variable shape compatibility for Components and Function-Parameters

kmantel:
Previously, only 'function' would have the correct shape. Now, all Functions that are Parameters can have their default variable and value set at instantiation time. Their owners' variable will be used, and passed through a _parse_<parameter name>_variable method if it exists on the owner.

mechanisms
----------
Use integrator_mode to determine integration status

Previously, this was done with checking for integrator_function being initialized, but this is now always the case, to ensure valid variable shape

Add argument 'force' to Mechanism.reset to ignore integrator_mode and proceed with resetting the integrator_function (it will always exist even when integration_mode is False)

RecurrentTransferMechanism.reset suppressed reset when integrator_mode was False, to prevent errors in Contrastive Hebbian. Replace this with a check in Contrastive Hebbian

tests
-----
Update numeric changes due to rng offset changes

Compared to devel, these tests changed as a result of the following numbers of calls to np.random methods during initialization:

| Test Name | numpy RNG method | offset |
| - | - | - |
test_transfer_mech_2d_variable_noise | np.random.normal | +1

Compared to devel, these tests changed as a result of the following numbers of calls to the random_state parameter of noise functions during initialization:

| Test Name | numpy RNG method | offset |
| - | - | - |
test_execute | random_state.normal | +1
test_accumulator_standalone_noise_function | random_state.normal | +1
test_integrator_simple_noise_fn | random_state.normal | +1
test_integrator_simple_noise_fn_var_list | random_state.normal | +1
test_integrator_accumulator_noise_fn | random_state.normal | +1
test_integrator_accumulator_noise_fn_var_list | random_state.normal | +1
test_integrator_adaptive_noise_fn | random_state.normal | +1
test_integrator_adaptive_noise_fn_var_list | random_state.normal | +1
test_buffer_standalone_noise_function | random_state.normal | +2

| Test Name | random_state Parameter RNG method | offset |
| - | - | - |
test_transfer_mech_array_var_normal_array_noise | random_state.normal | -1 (for each NormalDist instance)
test_transfer_mech_normal_noise | random_state.normal | -7
test_transfer_mech_exponential_noise | random_state.random | -7
test_transfer_mech_Uniform_noise | random_state.random | -7
test_transfer_mech_Gamma_noise | random_state.random | -7
test_transfer_mech_wald_noise | random_state.wald | -7
test_transfer_mech_array_var_normal_len_1_noise | random_state.normal | -7

### DIFF
--- a/psyneulink/core/components/component.py
+++ b/psyneulink/core/components/component.py
@@ -2654,6 +2654,20 @@ class Component(JSONDumpable, metaclass=ComponentsMeta):
 
     def _update_default_variable(self, new_default_variable, context=None):
         self.defaults.variable = copy.deepcopy(new_default_variable)
+
+        # exclude value from validation because it isn't updated until
+        # _instantiate_value is called
+        call_with_pruned_args(
+            self._validate_params,
+            variable=new_default_variable,
+            request_set={
+                k: v.default_value
+                for k, v in self.parameters.values(True).items()
+                if k not in {'value'} and not isinstance(v, ParameterAlias)
+            },
+            target_set={},
+            context=context
+        )
         self._instantiate_value(context)
 
         function_variable = self._parse_function_variable(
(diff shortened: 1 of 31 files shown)


kmantel:
This also fixes the general case of #1479

PR #1670, opened 2020-06-13
tests: Remove warning ambiguity in TestDuplicatePathwayWarnings

jvesely:
The two tested lines trigger 2 warnings which are captured in undefined
ordered.
It's the second line that triggers the warning were interested in so
move the first line out of the capture block.
Split 'and' in asserts while at it.


jvesely:
This fixes intermittent failures of the modified tests.

PR #1679, opened 2020-06-16
llvm, functions/UserDefinedFunction: Add python callback implementation of UDF

jvesely:


SamKG:
I'm curious as to how this would interact with the GIL in python - does invoking the python callback reacquire the lock?

jvesely:
> I'm curious as to how this would interact with the GIL in python - does invoking the python callback reacquire the lock?

yes, afaik it acquires the lock before executing any python code and releases it after.

PR #1685, opened 2020-06-20
tests: Cleanup

jvesely:
Cleanup uses of benchmark fixture and pytest.warns.


jvesely:
This PR also fixes test failures in test_memory when using numpy-1.19

PR #1698, opened 2020-07-05
llvm/cuda: Cleanup

jvesely:
Drop references to per node PTX mode.
Fix builtin random tests failures since configurable cuda grid block size.
Improve generation of input structures.


jvesely:
cc @SamKG

PR #1713, opened 2020-07-15
Reduce duration of model tests

kmantel:
Reduce the number of initialization/settlement trials, and relax convergence thresholds. Shortened test results still resemble the general trends of their full equivalents, but a standard test run of `tests/models/test_documentation_models.py` takes about 70% less time.

The original tests are marked as `stress`


tests/models/test_documentation_models.py
```diff
@@ -22,19 +22,58 @@
...
+        pytest.param('MontagueDayanSejnowski96', 'comp_5c', ['--figure', '5c'], False),
+        pytest.param('Nieuwenhuis2005Model', 'task', [], False),
     ]
 )
-def test_documentation_models(model_name, composition_name, additional_args):
+def test_documentation_models(
+    model_name,
+    composition_name,
+    additional_args,
+    reduced,
```
jvesely, line 75: I'd be more extensible to call the last arg `variant-suffix` or similar and pass 'reduced',
but I don't expect anyone to add more variants any time soon so it doesn't really matter.

PR #1739, opened 2020-08-11
tests/parameter_estimation: Add GridSearch version of the floating average test

jvesely:


psyneulink/core/compositions/composition.py
```diff
@@ -7335,7 +7335,8 @@ def _get_total_cost_of_control_allocation(self, control_allocation, context, run
                                                                 )
 
             # Get control signal costs
-            all_costs = convert_to_np_array(self.controller.parameters.costs._get(context) + [reconfiguration_cost])
+            other_costs = self.controller.parameters.costs._get(context) or []
```
kmantel, line 7338: At first, thought this could hide a case where `self.controller.parameters.costs._get(context)` returns 0, but `_control_mechanism_costs_getter` always returns `list` or `None` so it's fine.

jvesely: Thanks!

PR #1749, opened 2020-09-09
Initialize cycles

dillontsmith:
RPC

- add protobufs for loggable parameters and component lists

Composition

- add `Composition.initialize` helper method


jvesely:
Looks like the windows failures with `Con.proto` are were introduced in `grpcio-1.32.0`

I've reported a bug https://github.com/grpc/grpc/issues/24106, but maybe we can restrict the dependency on windows.

dillontsmith:
> Looks like the windows failures with `Con.proto` are were introduced in `grpcio-1.32.0`
> 
> I've reported a bug [grpc/grpc#24106](https://github.com/grpc/grpc/issues/24106), but maybe we can restrict the dependency on windows.

Thanks for the heads up.

jvesely:
> > Looks like the windows failures with `Con.proto` are were introduced in `grpcio-1.32.0`
> > I've reported a bug [grpc/grpc#24106](https://github.com/grpc/grpc/issues/24106), but maybe we can restrict the dependency on windows.
> 
> Thank you! Just pushed [a2cb01e](https://github.com/PrincetonUniversity/PsyNeuLink/commit/a2cb01e634af1f4953668ce9ddea9fa35d448257), which restricts `grpcio` to versions previous to `1.32.0`.

you probably need to restrict `grpcio-tools` as well, since they need to be in sync and `pip` is not good enough (yet) to handle that.

dillontsmith:
> > > Looks like the windows failures with `Con.proto` are were introduced in `grpcio-1.32.0`
> > > I've reported a bug [grpc/grpc#24106](https://github.com/grpc/grpc/issues/24106), but maybe we can restrict the dependency on windows.
> > 
> > 
> > Thank you! Just pushed [a2cb01e](https://github.com/PrincetonUniversity/PsyNeuLink/commit/a2cb01e634af1f4953668ce9ddea9fa35d448257), which restricts `grpcio` to versions previous to `1.32.0`.
> 
> you probably need to restrict `grpcio-tools` as well, since they need to be in sync and `pip` is not good enough (yet) to handle that.

Good call... 6b0810e restricts `grpcio-tools` as well.

PR #1759, opened 2020-09-22
llvm/cleanup/remove bitcasts

SamKG:
Removes bitcasts from various lllvm pytorch helpers


jvesely (approved):
thanks! it might be worth fixing the allocas (use `alloca(m1.type.pointee` ...) while here, but it can also be left for future cleanup.
alias analysis should have easier time handling out of bounds accesses than pointer casts.
we should still check that any data structure that we use doesn't insert extra padding.
It's probably not going to happen with fp64, but fp16 arrays can be padded/aligned to 4B

PR #1801, opened 2020-10-24
Context: generate an execution_id on creation if unspecified

kmantel:
- Solves an issue where manually creating a Context object could cause unintended use of the None or a default context.

- Adds arguments `fallback_default` and `fallback_most_recent` to the `handle_external_context` decorator that can be used to determine whether the function should use an object's `default_execution_id` or `most_recent_context.execution_id` if context is unspecified

### DIFF
--- a/tests/composition/test_composition.py
+++ b/tests/composition/test_composition.py
@@ -3764,6 +3764,21 @@ class TestRun:
             context='custom'
         )
 
+    def test_manual_context(self):
+        t = pnl.TransferMechanism()
+        comp = pnl.Composition()
+
+        comp.add_node(t)
+
+        comp.run({t: [1]})
+        assert comp.results == [[[1]]]
+
+        context = pnl.Context()
+        t.function.parameters.slope._set(2, context)
+
+        comp.run({t: [1]}, context=context)
+        assert comp.results == [[[2]]]
+
 
 class TestCallBeforeAfterTimescale:
 
@@ -6294,7 +6309,7 @@ class TestInitialize:
         C = RecurrentTransferMechanism(name='C',
                                        auto=1.0)
 
-        context = Context(execution_id='a') if context_specified else NotImplemented
+        context = Context(execution_id='a') if context_specified else None
 
         abc_Composition = Composition(pathways=[[A, B, C]])
 
@@ -6303,7 +6318,7 @@ class TestInitialize:
         abc_Composition.run(inputs={A: [1.0, 2.0, 3.0]}, context=context)
 
         if not context_specified:
-            abc_Composition.run()
+            abc_Composition.run(context=Context(execution_id='b'))
 
         abc_Composition.run(inputs={A: [1.0, 2.0, 3.0]}, context=context)
 
--- a/psyneulink/core/components/component.py
+++ b/psyneulink/core/components/component.py
@@ -1089,6 +1089,7 @@ class Component(JSONDumpable, metaclass=ComponentsMeta):
         context = Context(
             source=ContextFlags.COMPONENT,
             execution_phase=ContextFlags.IDLE,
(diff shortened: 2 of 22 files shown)


tests/composition/test_composition.py
```diff
@@ -6294,17 +6309,13 @@ def test_initialize_cycles(self, context_specified):
...
         abc_Composition.initialize({C: 2.0}, context=context)
 
         abc_Composition.run(inputs={A: [1.0, 2.0, 3.0]}, context=context)
-
-        if not context_specified:
-            abc_Composition.run()
-
         abc_Composition.run(inputs={A: [1.0, 2.0, 3.0]}, context=context)
 
         # Run 1 --> Execution 1: 1 + 2 = 3    |    Execution 2: 3 + 2 = 5    |    Execution 3: 5 + 3 = 8
```
kmantel, line 6321: before this PR, when `context_specified = False`, these execution_ids were used:

| | |
| - | - |
 initialize | None 
 run 1 | NotImplemented 
 run 2 | Composition-0 
 run 3 | NotImplemented 

The results ignored the second run because `abc_Composition.results` referred to the `NotImplemeted` context. My impression is that if context isn't specified, we'd want to use the default for the Composition (`"Composition-0"), so this PR would result in these execution_ids:

| | |
| - | - |
 initialize | None 
 run 1 | Composition-0
 run 2 | Composition-0 
 run 3 | Composition-0

But this would add an extra `[8]` to the results list due to run 2. I'm not sure if removing this extra run is against the purpose of the original test though, @dillontsmith

dillontsmith: Hi @kmantel,

I think perhaps the more descriptive parameter would have just been `dual_context`, for which a value of `True` would be equivalent to `context_specified` equaling `False`. The idea was just to have two separate contexts under that condition to make sure there weren't any nasty side effects for context switching that arose from use of the `initialize` method. The way I originally wrote it was just a consequence of how the defaults worked at the time, but I probably should have set it up to explicitly use two manually specified contexts from the outset.

Now that I'm looking at this, it might also make sense to call the test `test_initialize_method` or something to that effect, to distinguish it from the `initialize_cycle_values` arg of run, which is what is tested by the tests immediately before it.

kmantel: Ok, got it. So would
```diff
     @pytest.mark.parametrize("context_specified", [True, False])
     def test_initialize_cycles(self, context_specified):
         A = TransferMechanism(name='A')
         B = TransferMechanism(name='B')
         C = RecurrentTransferMechanism(name='C',
                                        auto=1.0)
 
-        context = Context(execution_id='a') if context_specified else NotImplemented
+        context = Context(execution_id='a') if context_specified else None
 
         abc_Composition = Composition(pathways=[[A, B, C]])
 
         abc_Composition.initialize({C: 2.0}, context=context)
 
         abc_Composition.run(inputs={A: [1.0, 2.0, 3.0]}, context=context)
 
         if not context_specified:
-            abc_Composition.run()
+            abc_Composition.run(context=Context(execution_id='b'))
 
         abc_Composition.run(inputs={A: [1.0, 2.0, 3.0]}, context=context)
 
         # Run 1 --> Execution 1: 1 + 2 = 3    |    Execution 2: 3 + 2 = 5    |    Execution 3: 5 + 3 = 8
         # Run 2 --> Execution 1: 8 + 1 = 9    |    Execution 2: 9 + 2 = 11    |    Execution 3: 11 + 3 = 14
         assert abc_Composition.results == [[[3]], [[5]], [[8]], [[9]], [[11]], [[14]]]
```

be faithful to the original intent?

dillontsmith: Yes, exactly.

PR #1835, opened 2020-12-01
deps: Add leabra as dependency to PsyNeuLink

dillontsmith:
Add `leabra` to `requirements.txt` and remove special casing from before the package was on PyPi

### DIFF
--- a/.github/workflows/pnl-ci.yml
+++ b/.github/workflows/pnl-ci.yml
@@ -74,7 +74,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip wheel
         pip install -e .[dev]
-        pip install --user git+https://github.com/benureau/leabra.git@master
 
     - name: Cleanup old wheels
       shell: bash
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,8 +114,6 @@ before_cache:
 install:
   - pip install coveralls
   - pip install -e .[dev]
-  # This should be installed after PNL to avoid pulling incompatible deps
-  - pip install git+https://github.com/benureau/leabra.git@master
 
 
 script:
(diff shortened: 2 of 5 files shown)


jvesely:
This needs CI changes similar to #1815

dillontsmith:
@jvesely, I was under the impression that the CI changes from #1815 were needed because the leabra requirement was 1) a direct dependency and 2) in a separate requirements file. Is this not the case?

dillontsmith:
retest this please

jvesely:
> @jvesely, I was under the impression that the CI changes from #1815 were needed because the leabra requirement was 1) a direct dependency and 2) in a separate requirements file. Is this not the case?

My understanding was that leabra-psyneulink replaces git+https://github.com/benureau/leabra.git@master
if so, 1.) should still be applied here. if leabra is pulled via a dependency file, the direct pip install from git should be removed from CI setup.

PS: I'd expect leabra-psyneulink and leabra.git to conflict on install since they both add 'leabra' module. wonder how both of them coexist.

dillontsmith:
Ah, I see. You're right, on this pull the CI is redundantly installing the original leabra package + the one on PyPi. Will upload a patch to fix.

jvesely:
> Ah, I see. You're right, on this pull the CI is redundantly installing the original leabra package + the one on PyPi. Will upload a patch to fix.
> 
> Edit: patched in [7a161e6](https://github.com/PrincetonUniversity/PsyNeuLink/commit/7a161e619307a6bc159b8500995596efadee7a05)

thanks. LGTM

dillontsmith:
retest this please

dillontsmith:
retest this please

PR #1837, opened 2020-12-01
Llvm/tyler ddm

SamKG:
add functionality to compiled udfs

### DIFF
--- a/psyneulink/core/llvm/codegen.py
+++ b/psyneulink/core/llvm/codegen.py
@@ -71,7 +71,13 @@ class UserDefinedFunctionVisitor(ast.NodeVisitor):
         # setup numpy
         numpy_handlers = {
             'tanh': _tanh,
-            'exp': _exp
+            'exp': _exp,
+            'equal': self._generate_fcmp_handler(self.ctx, self.builder, "=="),
+            'not_equal': self._generate_fcmp_handler(self.ctx, self.builder, "!="),
+            'less': self._generate_fcmp_handler(self.ctx, self.builder, "<"),
+            'less_equal': self._generate_fcmp_handler(self.ctx, self.builder, "<="),
+            'greater': self._generate_fcmp_handler(self.ctx, self.builder, ">"),
+            'greater_equal': self._generate_fcmp_handler(self.ctx, self.builder, ">="),
         }
 
         for k, v in func_globals.items():
@@ -392,53 +398,68 @@ class UserDefinedFunctionVisitor(ast.NodeVisitor):
 
         return _or
 
-    def visit_Eq(self, node):
-        def _eq(x, y):
-            assert helpers.is_floating_point(x), f"{x} is not a floating point type!"
-            assert helpers.is_floating_point(y), f"{y} is not a floating point type!"
-            return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered('==', x, y))
+    def _generate_fcmp_handler(self, ctx, builder, cmp):
+        def _cmp_array(ctx, builder, u, v):
+            assert u.type == v.type
+            shape = helpers.get_array_shape(u)
+            output_ptr = builder.alloca(helpers.array_from_shape(shape, ctx.bool_ty))
 
-        return _eq
+            for (u_ptr, v_ptr, out_ptr) in helpers.recursive_iterate_arrays(ctx, builder, u, v, output_ptr):
+                u_val = builder.load(u_ptr)
+                v_val = builder.load(v_ptr)
+                builder.store(builder.fcmp_ordered(cmp, u_val, v_val), out_ptr)
+
+            return output_ptr
+
+        def _cmp_array_scalar(ctx, builder, array, s):
+            shape = helpers.get_array_shape(array)
+            output_ptr = builder.alloca(helpers.array_from_shape(shape, ctx.bool_ty))
+            helpers.call_elementwise_operation(ctx, builder, array, lambda ctx, builder, x:  builder.fcmp_ordered(cmp, x, s), output_ptr)
+
+            return output_ptr
+
+        def _cmp_scalar_array(ctx, builder, s, array):
+            shape = helpers.get_array_shape(array)
+            output_ptr = builder.alloca(helpers.array_from_shape(shape, ctx.bool_ty))
+            helpers.call_elementwise_operation(ctx, builder, array, lambda ctx, builder, x:  builder.fcmp_ordered(cmp, s, x), output_ptr)
+
+            return output_ptr
+
+        def _cmp(x, y):
+            if helpers.is_floating_point(x) and helpers.is_floating_point(y):
+                return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered(cmp, x, y))
+            elif helpers.is_vector(x) and helpers.is_floating_point(y):
+                return self._generate_binop(x, y, _cmp_array_scalar)
+            elif helpers.is_floating_point(x) and helpers.is_vector(y):
+                return self._generate_binop(x, y, _cmp_scalar_array)
+            elif helpers.is_2d_matrix(x) and helpers.is_floating_point(y):
+                return self._generate_binop(x, y, _cmp_array_scalar)
+            elif helpers.is_floating_point(x) and helpers.is_2d_matrix(y):
+                return self._generate_binop(x, y, _cmp_scalar_array)
+            elif helpers.is_vector(x) and helpers.is_vector(y):
+                return self._generate_binop(x, y, _cmp_array)
+            elif helpers.is_2d_matrix(x) and helpers.is_2d_matrix(y):
+                return self._generate_binop(x, y, _cmp_array)
+
+        return _cmp
+
+    def visit_Eq(self, node):
+        return self._generate_fcmp_handler(self.ctx, self.builder, "==")
 
     def visit_NotEq(self, node):
-        def _neq(x, y):
-            assert helpers.is_floating_point(x), f"{x} is not a floating point type!"
-            assert helpers.is_floating_point(y), f"{y} is not a floating point type!"
-            return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered('!=', x, y))
-
-        return _neq
+        return self._generate_fcmp_handler(self.ctx, self.builder, "!=")
 
     def visit_Lt(self, node):
-        def _lt(x, y):
-            assert helpers.is_floating_point(x), f"{x} is not a floating point type!"
-            assert helpers.is_floating_point(y), f"{y} is not a floating point type!"
-            return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered('<', x, y))
-
-        return _lt
+        return self._generate_fcmp_handler(self.ctx, self.builder, "<")
 
     def visit_LtE(self, node):
-        def _lte(x, y):
-            assert helpers.is_floating_point(x), f"{x} is not a floating point type!"
-            assert helpers.is_floating_point(y), f"{y} is not a floating point type!"
-            return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered('<=', x, y))
-
-        return _lte
+        return self._generate_fcmp_handler(self.ctx, self.builder, "<=")
 
     def visit_Gt(self, node):
-        def _gt(x, y):
-            assert helpers.is_floating_point(x), f"{x} is not a floating point type!"
-            assert helpers.is_floating_point(y), f"{y} is not a floating point type!"
-            return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered('>', x, y))
-
-        return _gt
+        return self._generate_fcmp_handler(self.ctx, self.builder, ">")
 
     def visit_GtE(self, node):
-        def _gte(x, y):
-            assert helpers.is_floating_point(x), f"{x} is not a floating point type!"
-            assert helpers.is_floating_point(y), f"{y} is not a floating point type!"
-            return self._generate_binop(x, y, lambda ctx, builder, x, y: builder.fcmp_ordered('>=', x, y))
-
-        return _gte
+        return self._generate_fcmp_handler(self.ctx, self.builder, ">=")
 
     def visit_Compare(self, node):
         comp_val = self.visit(node.left)
--- a/psyneulink/core/llvm/codegen.py
+++ b/psyneulink/core/llvm/codegen.py
@@ -264,6 +264,11 @@ class UserDefinedFunctionVisitor(ast.NodeVisitor):
             elif helpers.is_floating_point(x) and helpers.is_2d_matrix(y):
                 return self._generate_binop(y, x, _mul_mat_scalar)
             elif helpers.is_vector(x) and helpers.is_vector(y):
+                if x.type != y.type:
+                    # Special case: Cast y into scalar if it can be done
+                    if helpers.get_array_shape(y) == [1]:
+                        y = self.builder.gep(y, [self.ctx.int32_ty(0), self.ctx.int32_ty(0)])
+                        return self._generate_binop(x, y, _mul_vec_scalar)
                 return self._generate_binop(x, y, _mul_vec)
             elif helpers.is_2d_matrix(x) and helpers.is_2d_matrix(y):
                 return self._generate_binop(x, y, _mul_mat)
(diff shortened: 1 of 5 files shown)


dillontsmith:
retest this please

jvesely (changes requested):
The first round of reviews.

psyneulink/core/llvm/codegen.py
```diff
@@ -260,6 +350,44 @@ def visit_Name(self, node):
...
+        elif node.attr == "astype":
+            def astype(ty):
+                def _convert(ctx, builder, x):
+                    if helpers.is_pointer(x):
+                        x = builder.load(x)
+                    if helpers.is_integer(x) and ty is ctx.float_ty:
+                        if helpers.is_boolean(x):
+                            return builder.uitofp(x, ty)
+                        return builder.sitofp(x, ty)
+                    elif helpers.is_floating_point(x) and ty is self.ctx.float_ty:
```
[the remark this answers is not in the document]
SamKG: this case (float -> int) actually turns out to be difficult to test for, since return types from compiled udfs are always assumed to be doubles.
in addition, the way I detect size of the integers might result in different int sizes than the rest of the compiled code

it may be worth it to not allow float -> int conversions altogether if this is the case

psyneulink/core/llvm/helpers.py
```diff
@@ -233,12 +222,54 @@ def is_2d_matrix(x):
...
+
+def array_from_shape(shape, element_ty):
+    array_ty = element_ty
+    for dim in reversed(shape):
+        array_ty = ir.ArrayType(array_ty, dim)
+    return array_ty
+
+def recursive_iterate_array(ctx, builder, u, *args):
+    """Recursively iterates over all elements in scalar arrays of the same shape"""
+    assert all(get_array_shape(u) == get_array_shape(v) for v in args), "Tried to iterate over differing types!"
```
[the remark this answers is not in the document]
SamKG: the method should be able to iterate over any arrays of the same shape - including those with differing types (see `codegen:_generate_fcmp_handler` for an example of where this is used - in essence, the output array of booleans is populated while iterating)

if this is an anti-pattern, please do let me know.

psyneulink/core/llvm/codegen.py
```diff
@@ -260,6 +350,44 @@ def visit_Name(self, node):
...
+        if node.attr == "shape":
+            # recursively extract shape from value
+            dimensions = []
+            val = val.type.pointee
+            while hasattr(val, "count"):
+                dimensions.append(val.count)
+                val = val.element
+
+            # dimensions is a tuple, so it should be a llvm struct
+            dimensions_struct = self.builder.alloca(ir.LiteralStructType([self.ctx.float_ty] * len(dimensions)))
```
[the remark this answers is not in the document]
SamKG: this is a struct of floats to match the python return type (tuple), as tuples are interpreted as structs in the compiled environment.

this could technically be fixed by converting arrays to structs and structs to arrays upon return (to explicitly cast it to the expected types). I am not sure if this would have unintended side-effects, however.

psyneulink/core/llvm/codegen.py
```diff
@@ -260,6 +350,44 @@ def visit_Name(self, node):
...
+            for idx, dim in enumerate(dimensions):
+                struct_ptr = self.builder.gep(dimensions_struct, [self.ctx.int32_ty(0), self.ctx.int32_ty(idx)])
+                self.builder.store(self.ctx.float_ty(dim), struct_ptr)
+            return dimensions_struct
+        elif node.attr == "astype":
+            def astype(ty):
+                def _convert(ctx, builder, x):
+                    if helpers.is_pointer(x):
+                        x = builder.load(x)
+                    if helpers.is_integer(x) and ty is ctx.float_ty:
```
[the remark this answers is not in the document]
SamKG: the ctx parameter was added for the `call_elementwise_operation` function.
in hindsight, it's probably possible to completely replace `call_elementwise_operation` with `recursive_iterate_arrays` and remove any explicit parameter requirements from the callback.

psyneulink/core/llvm/helpers.py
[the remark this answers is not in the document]
jvesely: so why can't you compare the types directly?

psyneulink/core/llvm/codegen.py
```diff
@@ -260,6 +344,36 @@ def visit_Name(self, node):
...
+                        if helpers.is_boolean(x):
+                            return builder.uitofp(x, ty)
+                        return builder.sitofp(x, ty)
+                    elif helpers.is_floating_point(x) and ty is self.register["int"]:
+                        return builder.fptosi(x, ty)
+                    elif (helpers.is_floating_point(x) and ty is ctx.float_ty) or \
+                         (helpers.is_integer(x) and ty is self.register["int"]):
+                        return x
+                    else:
+                        raise Exception(f"Tried to convert {x} to invalid type: {ty}")
```
jvesely, line 367: No runtime exceptions. They indicate error to the user, there's nothing in the compiler that the user should care about.

jvesely (changes requested):

psyneulink/core/llvm/codegen.py
```diff
@@ -49,12 +61,39 @@ def _exp(x):
             helpers.call_elementwise_operation(self.ctx, self.builder, x, helpers.exp, output_ptr)
             return output_ptr
 
-        self.register['sum'] = _vec_sum
+        def _max(x):
+            assert helpers.is_vector(x) or helpers.is_2d_matrix(x), "Attempted to call max on invalid variable! Only 1-d and 2-d lists are supported!"
+            curr = builder.alloca(ctx.float_ty)
+            has_set = False
```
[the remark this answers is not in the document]
SamKG: this requires the condition to be 
`greater = builder.not_(builder.fcmp_ordered('<', element, builder.load(curr)))`,
since ordered comparisons always return `False` is one of the operands is `NaN`.

Fixed by swapping to unordered comparison.

psyneulink/core/llvm/codegen.py
```diff
@@ -28,16 +28,28 @@ def __init__(self, ctx, builder, func_globals, func_params, arg_in, arg_out):
...
-            dim = len(x.type.pointee)
-            output_scalar = builder.alloca(ctx.float_ty)
-            # Get the pointer to the first element of the array to convert from [? x double]* -> double*
-            vec_u = builder.gep(x, [ctx.int32_ty(0), ctx.int32_ty(0)])
-            builder.call(ctx.import_llvm_function("__pnl_builtin_vec_sum"), [vec_u, ctx.int32_ty(dim), output_scalar])
-            return output_scalar
+        def _list_sum(x):
+            # HACK: Obtain generic addition function by visiting the add node
+            # this should ideally be moved to an explicit helper
+            add_func = self.visit_Add(None)
```
jvesely, line 36: The comment is not quite clear here. Does the `visit_Add` here return the right addition for different element types (float/int/array)?

psyneulink/core/llvm/helpers.py
```diff
@@ -239,6 +228,42 @@ def is_boolean(x):
...
+def array_from_shape(shape, element_ty):
+    array_ty = element_ty
+    for dim in reversed(shape):
+        array_ty = ir.ArrayType(array_ty, dim)
+    return array_ty
+
+def recursive_iterate_arrays(ctx, builder, u, *args):
+    """Recursively iterates over all elements in scalar arrays of the same shape"""
+    assert isinstance(u.type.pointee, ir.ArrayType), "Can only iterate over arrays!"
+    assert all(len(u.type.pointee) == len(v.type.pointee) for v in args), "Tried to iterate over differing types!"
```
jvesely, line 253: nit: the assertion message should indicate that the length was different rather than type

psyneulink/core/llvm/helpers.py
```diff
@@ -239,6 +228,42 @@ def is_boolean(x):
...
+    with array_ptr_loop(builder, u, str(u) + "," + str(args) + "_recursive_zip") as (b, idx):
+        u_ptr = b.gep(u, [ctx.int32_ty(0), idx])
+        arg_ptrs = (b.gep(v, [ctx.int32_ty(0), idx]) for v in args)
+        if is_scalar(u_ptr):
+            yield (u_ptr, *arg_ptrs)
+        else:
+            yield from recursive_iterate_arrays(ctx, b, u_ptr, *arg_ptrs)
+
+def call_elementwise_operation(ctx, builder, x, operation, output_ptr):
```
jvesely, line 262: This can be marked for removal. It doesn't really add anything on top of the iterator and leads to overly long lines. At that point splitting the invocation into two lines is more readable anyway.
But that is for later cleanup (not in this PR)

tests/functions/test_user_defined_func.py
```diff
@@ -109,80 +140,67 @@ def myFunction(variable):
...
+                        ("GtE", 1.0, [1.0, 2.0], [1.0, 0.0]),
+                        ("GtE", [2.0, 1.0], 1.0, [1.0, 1.0]),
                         ])
     @pytest.mark.parametrize("bin_execute", ['Python',
                                              pytest.param('LLVM', marks=pytest.mark.llvm),
                                              pytest.param('PTX', marks=[pytest.mark.llvm, pytest.mark.cuda]),
                                             ])
     @pytest.mark.benchmark(group="Function UDF")
-    def test_user_def_func_cmpop(self, op, bin_execute, benchmark):
+    def test_user_def_func_cmpop(self, op, var1, var2, expected, bin_execute, benchmark):
```
jvesely, line 174: This used to test Pyhton comparison operators, but was switched to numpy.
It also combines testing of comparison an `as_type` in one tests.
Replacing the lines with `1.0 if np.op(var1, var2) else 0.0` would be more directed.

the ops should also use space after comman consistently.

tests/llvm/test_helpers.py
```diff
@@ -519,6 +519,48 @@ def test_helper_is_boolean(self, mode, ir_type, expected):
...
+        (DOUBLE_VECTOR_TYPE, (1,)),
+        (DOUBLE_VECTOR_PTR_TYPE, (1,)),
+        (DOUBLE_MATRIX_TYPE, (1,1)),
+        (DOUBLE_MATRIX_PTR_TYPE, (1,1)),
+    ], ids=str)
+    def test_helper_get_array_shape(self, mode, ir_type, expected):
+        with pnlvm.LLVMBuilderContext() as ctx:
+            func_ty = ir.FunctionType(ir.VoidType(), [ir.IntType(32).as_pointer()])
+
+            custom_name = ctx.get_unique_name("is_boolean")
```
jvesely, line 535: probably shouldn't be called 'is_boolean'

tests/llvm/test_helpers.py
```diff
@@ -519,6 +519,48 @@ def test_helper_is_boolean(self, mode, ir_type, expected):
...
+            function = ir.Function(ctx.module, func_ty, name=custom_name)
+            out = function.args[0]
+            block = function.append_basic_block(name="entry")
+            builder = ir.IRBuilder(block)
+
+            variable = builder.load(builder.alloca(ir_type))
+            assert pnlvm.helpers.get_array_shape(variable) == expected
```
jvesely, line 542: this should need compilation at all.
`assert pnlvm.helpers.get_array_shape(ir_type(None)) == expected` should do fine

tests/functions/test_user_defined_func.py
```diff
@@ -318,25 +355,38 @@ def myFunction(variable):
...
                 return np.tanh(variable)
         elif op == "EXP":
             def myFunction(variable):
                 return np.exp(variable)
+        elif op == "SHAPE":
+            def myFunction(variable):
+                return variable.shape
+        elif op == "ASTYPE_FLOAT":
+            def myFunction(variable):
+                return variable.astype(float)
```
jvesely, line 383: Can input types be integer? or does this suffer from the same issue as the variant below?

psyneulink/core/llvm/helpers.py
```diff
@@ -239,6 +228,42 @@ def is_boolean(x):
...
+    if is_pointer(x):
+        x_ty = x_ty.pointee
+
+    assert isinstance(x_ty, ir.ArrayType), f"Tried to get shape of non-array type: {x_ty}"
+    dimensions = []
+    while hasattr(x_ty, "count"):
+        dimensions.append(x_ty.count)
+        x_ty = x_ty.element
+
+    return tuple(dimensions)
```
jvesely, line 242: Why does this need to return tuple?

jvesely (commented):
Few more nits.
Is coveralls happy (all new code covered by tests)?

psyneulink/core/llvm/codegen.py
SamKG: yes - this is the polymorphic addition function as used when visiting an addition node

tests/functions/test_user_defined_func.py
SamKG: > This used to test Pyhton comparison operators, but was switched to numpy.
Good catch - I'll split the tests so there is one for numpy and one for normal comparisons

Replacing with `1.0 if np.op(var1, var2) else 0.0` isn't possible, as numpy comparison ops return an array where each op is done elementwise (e.g. `np.equal([1,0], [1,2])` returns `[True, False]`). The astype call here is needed to convert the resulting boolean array into something that can be returned from compiled mode.

psyneulink/core/llvm/helpers.py
SamKG: this is to match the type of numpy's shape attribute, which is a tuple. This probably isn't strictly necessary, but I didn't see a reason to allow for mutable shapes and to deviate from numpy's type.

tests/functions/test_user_defined_func.py
SamKG: indeed - this suffers the same issue, where the input type can only be float.

jvesely: so this version only tests the no-op float->float variant and the one below tests float->int->float.
I think int->int no-op should also be tested.

psyneulink/core/llvm/codegen.py
jvesely: yes, but you pass `None`, so what's the mechanism to infer the correct type? is that taken from the parser state?

psyneulink/core/llvm/helpers.py
jvesely: I don't think matching python in anything but computational semantics should be a priority.

psyneulink/core/llvm/codegen.py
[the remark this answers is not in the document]
jvesely: makes sense. we can replace it with array iteration now, or leave it for later cleanup since elementwise_op is used in other places. you call.

tests/functions/test_user_defined_func.py
SamKG: not sure how this can be done. is there a way you had in mind to do this?

psyneulink/core/llvm/helpers.py
SamKG: is the fix here to return the list without casting?

jvesely: yeah, unless something depends on it being a tuple. it might be easier to pass it as an initializer (ctypes likes tuples, dunno about llvmlite). that was the original question.
If the only reason is to stay close to Python, we can drop the cast.

tests/functions/test_user_defined_func.py
jvesely: makes sense, ifexp is better left for later

tests/functions/test_user_defined_func.py
jvesely: should we just drop the support for int->int casting then? if it's tricky to use, it might not be needed.

psyneulink/core/llvm/codegen.py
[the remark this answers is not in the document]
jvesely: representing tuples as structs is just a simplification and not a requirement
Python tuples allow dynamic indexing but struct type only works with constants so using structs is more restrictive.

psyneulink/core/llvm/codegen.py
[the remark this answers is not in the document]
jvesely: imo, it's good for now. it can be revisited later.

tests/functions/test_user_defined_func.py
SamKG: From a user standpoint, it would be extremely confusing to have to carefully manage types from within Python (e.g. udf returns int, and downstream, multiplies by a float. this would throw an opaque error with no clear fix).

It may be best to restrict udfs to only consider floats - that is, by casting any numbers that are integer constants into floats automatically (potentially warning the user if there are negative side-effects).

jvesely: Why would they need to 'carefully manage types' ? Either int->int casts are easy to have in UDF and should be tested, or it's an extreme case that is difficult to hit and can be dropped.

SamKG: removed int->int casts

jvesely (approved):

PR #1871, opened 2021-01-11
Add CI script checking differences in generated documentation

jvesely:
Don't run full CI on documentation only changes.
Adds script that generates 'before' and 'after' html documentation and posts diff as a comment.
Comment posting works only on PRs within one repo (should cover  dependabot requests)


jvesely:
Needs #1872  to land first for the new CI script to pass.

PR #1877, opened 2021-01-15
Consolidate and generalize noise value generation

kmantel:
`_try_execute_param`, which generates noise values from functions, is moved to Component and now supports variables of any length or dimension. Noise may be specified with the same flexibility and is applied elementwise, as in the `i`th element of noise will be executed (if a function) and applied to the `i`th element of variable, including any deeper dimension. For example, with a noise specification of `[pnl.NormalDist(), pnl.UniformDist()]` and a variable of `[[a, b], [c]]`, normal noise will be individually applied to elements `a` and `b` and uniform noise to `c`. 

There are two convenience exceptions:

1. if the variable shape is the same as the noise shape except with extra wrapping dimensions, the noise will be applied to the deepest dimension of variable (e.g. variable `[[[[a, b]]]]` and noise `[pnl.NormalDist(), pnl.UniformDist()]`, normal will apply to `a` and uniform to `b`. 
2. if noise is specified as a single psyneulink Function whose output/value shape matches variable, that noise function will be executed once and its value will be applied elementwise to variable

The following test expected values were changed because repeated executions of the tested mechanism incorrectly used the same numeric noise values as the first execution, which is now fixed:
    
- `test_accumulator_standalone_noise_function_in_array`
- `test_transfer_mech_array_var_normal_array_noise`

### DIFF
--- a/psyneulink/core/components/component.py
+++ b/psyneulink/core/components/component.py
@@ -520,7 +520,7 @@ from psyneulink.core.globals.registry import register_category
 from psyneulink.core.globals.utilities import \
     ContentAddressableList, convert_all_elements_to_np_array, convert_to_np_array, get_deepcopy_with_shared,\
     is_instance_or_subclass, is_matrix, iscompatible, kwCompatibilityLength, prune_unused_args, \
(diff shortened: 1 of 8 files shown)


kmantel:
retest this please

PR #1885, opened 2021-01-19
Feat/llvm/udf

SamKG:
adds support for np array flatten + addresses behavior differences between python and numpy max functions


jvesely (approved):
LGTM

PR #1886, opened 2021-01-19
Feat/convnet

SamKG:


psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -108,15 +148,15 @@ def log_value(self):
             self._mechanism.output_port.parameters.value._set(detached_value, self._context)
             self._mechanism.parameters.value._set(detached_value, self._context)
 
-    def _gen_llvm_execute_derivative_func(self, ctx, builder, state, params, arg_in):
+    def _update_llvm_param_gradients(self, ctx, builder, state, params, node_delta_w, z_value, backpropagated_error, *, tags=frozenset()):
```
jvesely, line 200: Why are you adding a `tags` argument? should the import below use `tags.union({"derivative"})?
The function signature as a whole does not match the parent prototype.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -132,24 +172,35 @@ def _gen_llvm_execute_derivative_func(self, ctx, builder, state, params, arg_in)
...
+        activation_func_derivative = builder.gep(output, [ctx.int32_ty(0),
+                                                          ctx.int32_ty(0)])
+
+        error_val = gen_inject_vec_hadamard(ctx, builder, activation_func_derivative, backpropagated_error)
+        return error_val
+
+    def _extract_llvm_param_ptr(self, ctx, builder, params, param_id):
+        mechanism_params = builder.gep(params, [ctx.int32_ty(0),
+                                           ctx.int32_ty(0),
+                                           ctx.int32_ty(self._idx)])
```
jvesely, line 233: Indentation looks wrong

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -163,6 +214,25 @@ def __init__(self, projection, component_idx, port_idx, device, sender=None, rec
 
         if projection.learnable is False:
             self.matrix.requires_grad = False
+    @property
```
jvesely, line 266: blank line between methods

psyneulink/library/compositions/pytorchmodelcreator.py
```diff
@@ -43,29 +46,34 @@ def __init__(self, composition, device, context=None):
...
+        for component in self.components:
+            for param in component._get_pytorch_params():
+                self.params.append(param)
 
         # Setup execution sets
         # 1) Remove all learning-specific nodes
         self.execution_sets = [x - set(composition.get_nodes_by_role(NodeRole.LEARNING)) for x in composition.scheduler.run(context=context)]
         # 2) Convert to pytorchcomponent representation
-        self.execution_sets = [{self.component_map[comp] for comp in s if comp in self.component_map} for s in self.execution_sets]
+        self.execution_sets = [{self.node_map[comp] for comp in s if comp in self.node_map} for s in self.execution_sets]
```
jvesely, line 76: Reassignment  makes things difficult to read each step should use a properly named temp variable.

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -37,11 +38,49 @@ def get_fct_param_value(param_name):
...
+    def _copy_params_to_psyneulink(self):
+        for param_id in self._get_learnable_param_ids():
+            pytorch_param = getattr(self, param_id)
+            param = getattr(self._object.parameters, param_id)
+
+            pytorch_value = pytorch_param.detach().cpu().numpy()
+            param._set(pytorch_value, self._context)
+
+            param_port = self._object.parameter_ports[param_id]
+            param_port.parameters.value._set(pytorch_value, self._context)
```
jvesely, line 39: Why is this setting ParameterPort `value` to the same value as the base param?
Shouldn't parameter value be after modulation?

psyneulink/library/compositions/pytorchcomponents.py
```diff
@@ -37,11 +38,49 @@ def get_fct_param_value(param_name):
...
+
+class PytorchWrapper():
+    def _get_learnable_param_ids(self):
+        return []
+
+    def _get_learnable_param_struct_type(self, ctx):
+        structs = []
+        for param_id in self._get_learnable_param_ids():
+            param = getattr(self._object.parameters, param_id)
+            assert param is not None, f"Failed to get param {param_id} from {self}"
```
jvesely, line 19: This assert message is deceptive. `gettattr` raises an exception if the attribute is present.
iiuc, `parameters.?` will get an instance of `Parameter` class and cannot be `None`.

jvesely (changes requested):
Few style nits and clarifications.

psyneulink/library/compositions/pytorchcomponents.py
SamKG: yes, the intent was to union the derivative tag. will fix

psyneulink/library/compositions/pytorchcomponents.py
SamKG: this is dead code - removed

psyneulink/library/compositions/pytorchcomponents.py
SamKG: actually, I think i can write this in a more general form (removing the long list of args) to begin with. I will take a stab at this.

tests/functions/test_transfer.py
```diff
@@ -135,6 +135,42 @@ def test_execute_derivative(func, variable, params, expected, benchmark, mode):
...
+    (np.ones((4, 3)), np.ones((1, 3)), (1,1), (0,0), (2,1)),
+    (np.ones((4, 3)), np.ones((1, 3)), (2,1), (0,0), (1,1)),
+    (np.ones((4, 3)), np.ones((1, 3)), (1,1), (1,1), (1,1)),
+])
+@pytest.mark.parametrize("mode", [
+    'Python',
+    pytest.param('LLVM', marks=pytest.mark.llvm),
+    pytest.param('PTX', marks=[pytest.mark.llvm, pytest.mark.cuda])])
+def test_conv2d_pytorch_equivalence(variable, kernel, stride, padding, dilation, benchmark, mode):
+    import torch
```
jvesely, line 155: This is not a good idea. no inline imports. torch equivalence should be tested separately.

The function also needs a traditional python/LLVM test.

psyneulink/library/compositions/pytorchcomponents.py
SamKG: nvm - this will require more work than previously thought.

for now, I have updated the `_update_llvm_param_gradients` function:
1) standardized signature
2) remove tags pass in (wasn't ever passed in anyways)
3) removed `*` from signature (shouldn't be necessary)

SamKG:
retest this please

tests/functions/test_transfer.py
```diff
@@ -135,6 +135,36 @@ def test_execute_derivative(func, variable, params, expected, benchmark, mode):
...
+    if mode == 'Python':
+        ex = f
+    elif mode == 'LLVM':
+        ex = pnlvm.execution.FuncExecution(f).execute
+    elif mode == 'PTX':
+        ex = pnlvm.execution.FuncExecution(f).cuda_execute
+    res = ex(variable)
+    assert np.allclose(res, expected)
+    if benchmark.enabled:
+        benchmark(f.function, variable)
```
jvesely, line 166: this always benchmarks the Python version.
if the function is not stateful it's better to use:
`res = benchmark(ex, variable)` 
above

jvesely (changes requested):
one nit and a test fix otherwise looks good

SamKG:
retest this please

PR #1896, opened 2021-01-28
llvm: cleanups and small fixes

jvesely:


SamKG (approved):

psyneulink/core/llvm/helpers.py
```diff
@@ -541,15 +533,7 @@ def generate_sched_condition(self, builder, condition, cond_ptr, node, is_finish
...
-            less_than_call_count = builder.icmp_unsigned('==', target_runs, self.ctx.int32_ty(count))
-
-            # Check that we have not run yet
-            my_time_stamp = self.__get_node_ts(builder, cond_ptr, node)
-            target_time_stamp = self.__get_node_ts(builder, cond_ptr, target)
-            ran_after_me = self.ts_compare(builder, my_time_stamp, target_time_stamp, '<')
-
-            # Return: target.calls % N == 0 AND me.last_time < target.last_time
-            return builder.and_(less_than_call_count, ran_after_me)
+            return builder.icmp_unsigned('==', target_runs, self.ctx.int32_ty(count))
```
SamKG, line 540: how exactly does target_runs work?
is it reset after every run of the composition?

jvesely: yes, the entire scheduler state gets thrown away after return from `run`.
the scheduler currently tracks number of executions in the current `run`, and the timestep (trial, pass, step) of the last execution.
There's a pretty big bug in how data dependencies are transformed into scheduler rules, and we probably need to use the full `execution_count` to properly implement most of the scheduling rules since a lot of them include `TimeScale` parameter.

Issue #1878, opened 2021-01-15
DriftDiffusionIntegrator always produces at least 3d output


kmantel:
```
>>> f = pnl.DriftDiffusionIntegrator()
>>> print(f.defaults.value)
(array([[0]]), array([0.]))
>>> print(f.execute(1))
(array([[1.]]), array([1.]))
>>> ddm = pnl.DDM(function=f)
[[[0.]]

 [[1.]]]
>>> print(ddm.defaults.value)
[[[1.0]]

 [[2.0]]]
```

I think there's a disconnect with what should be the correct value shape when we have multiple "stateful attributes" as we've called them (here, previous_value and previous_time). There are a couple lines that add dimensions - [integratorfunctions.py](https://github.com/PrincetonUniversity/PsyNeuLink/blob/b00a9b756428dac3241af4c66928e5319792f3d0/psyneulink/core/components/functions/statefulfunctions/integratorfunctions.py#L2469-L2473)
```
        if not np.isscalar(variable):
                previous_time = np.broadcast_to(
                    previous_time,
                    variable.shape
                ).copy()
```
and [ddm.py](https://github.com/PrincetonUniversity/PsyNeuLink/blob/029bcf79312aa2a01452f9af19552afee6cb9076/psyneulink/library/components/mechanisms/processing/integrator/ddm.py#L1050)
```
return convert_to_np_array([result[0], [result[1]]])
```

I think this brings up a problem with the shape of the stateful attributes that is masked when only previous_value is used. Most of the time, previous_value is cast to 2d or the shape of variable, which is often 2d when a function is associated with a mechanism and passed directly to the function. This would result in 3d output when multiple stateful attributes are returned. I think the automatic casting of previous_value should be reconsidered, but this currently causes many test failures.

kmantel:
Closed from #1899

PR #1899, opened 2021-02-02
DriftDiffusionIntegrator: avoid always producing 3d value

kmantel:
Fixes the specific case from #1878 but not in general. DriftDiffusionIntegrator and any owning mechanism now produces 2d value. Removes automatic 2d-cast of previous_value.


jvesely:
I'm not sure I follow the intention here. Don't all functions always return 2d output?

kmantel:
Not in general, for example 
```
>>> pnl.Linear().execute(5)
array([5.])

>>> pnl.DriftDiffusionIntegrator().execute(5)
(array([[5.]]), array([1.]))
```

when assigned to a DDM,
```
>>> pnl.DDM(function=pnl.DriftDiffusionIntegrator()).execute(5)
array([[[5.0]],

       [[1.0]]], dtype=object)
```

This isn't really a "full" solution to potential ragged arrays when returning multiple items from a function though. Mainly there were a few changes to llvm code that I wasn't sure were the best way of going about it.

jvesely:
the LLVM changes in this PR look good (in line with the rest of the changes).

I think it'd be more robust if DriftDiffusionIntegrator accepted 2d variables and just called variable.flatten() instead of mechanisms handling the conversion. DDI can theoretically be used as an integrator function in TransferMechanism, no?

jvesely (approved):

Issue #1802, opened 2020-10-24
Unstable code coverage for leabra


jvesely:
The leabra mechanism switches between covering fallback paths for `network` parameter.
Example covered: https://coveralls.io/builds/34318351/source?filename=psyneulink%2Flibrary%2Fcomponents%2Fmechanisms%2Fprocessing%2Fleabramechanism.py#L312
Example not-covered: https://coveralls.io/builds/34396708/source?filename=psyneulink%2Flibrary%2Fcomponents%2Fmechanisms%2Fprocessing%2Fleabramechanism.py#L312

jvesely:
The discrepancy is caused by non-determinism in initializing Parameters.
Setting the `training_flag` Parameter requires the `network` Parameter to be available.
Which in turn requires the `function` argument to be available.
If the above does not hold accessing `function.network` will raise `AttributeError` and take the exception handling paths

PR #1909, opened 2021-02-15
LeabraMechanism: standardize function creation, make network FunctionParameter

jvesely:
Closes: https://github.com/PrincetonUniversity/PsyNeuLink/issues/1802


jvesely:
@kmantel, this is your patch that we discussed to avoid non-determinism in leabra.
It doesn't solve the underlying priority issue, but it at least makes things deterministic, which is an improvement.

kmantel (approved):

kmantel:
Either way, these changes make LeabraMechanism more consistent with other mechanisms

PR #1912, opened 2021-02-16
scheduler: Use AllHaveRun(PASS) instead of EveryNCalls to fill in empty sched conditions

jvesely:

### DIFF
--- a/psyneulink/core/scheduling/scheduler.py
+++ b/psyneulink/core/scheduling/scheduler.py
@@ -283,7 +283,7 @@ from toposort import toposort
 
 from psyneulink.core.globals.context import Context, handle_external_context
 from psyneulink.core.globals.json import JSONDumpable
-from psyneulink.core.scheduling.condition import All, AllHaveRun, Always, Condition, ConditionSet, EveryNCalls, Never
+from psyneulink.core.scheduling.condition import AllHaveRun, Always, Condition, ConditionSet, Never
 from psyneulink.core.scheduling.time import Clock, TimeScale
 
 __all__ = [
@@ -574,10 +574,8 @@ class Scheduler(JSONDumpable):
 
                 if node_index > 0:
                     dependencies = list(self.consideration_queue[i - 1])
-                    if len(dependencies) == 1:
-                        cond = EveryNCalls(dependencies[0], 1)
-                    elif len(dependencies) > 1:
-                        cond = All(*[EveryNCalls(x, 1) for x in dependencies])
+                    if len(dependencies) > 0:
+                        cond = AllHaveRun(*dependencies, time_scale=TimeScale.PASS)
                     else:
                         raise SchedulerError(f'{self}: Empty consideration set in consideration_queue[{i - 1}]')
                 else:


jvesely:
hi @kmantel.
I'm not sure what the purpose of this code is. It appears to add a basic scheduling condition if a node has none.
but it's not added to every node, and it doesn't look to impact Python scheduling since the rules follow the consideration queue.
This change is analogous to 8f5f82c30f0a2d4a84fd66ea19ab53130e5814e9 ,but I kept the `All` part instead of switching to `Any`.

I'm not sure why compiled test cases don't hit this either. It only shows up in `Python-PTX` mode for GPU execution of controllers. probably not all the validation checks run before a composition is compiled, which might explain some of the shape mismatch errors.

kmantel:
> hi @kmantel.
> I'm not sure what the purpose of this code is. It appears to add a basic scheduling condition if a node has none.
> but it's not added to every node, and it doesn't look to impact Python scheduling since the rules follow the consideration queue.
> This change is analogous to [8f5f82c](https://github.com/PrincetonUniversity/PsyNeuLink/commit/8f5f82c30f0a2d4a84fd66ea19ab53130e5814e9) ,but I kept the `All` part instead of switching to `Any`.
> 
> I'm not sure why compiled test cases don't hit this either. It only shows up in `Python-PTX` mode for GPU execution of controllers. probably not all the validation checks run before a composition is compiled, which might explain some of the shape mismatch errors.
> 
> Fixes: Python-PTX failures since [dac19d4](https://github.com/PrincetonUniversity/PsyNeuLink/commit/dac19d4fb6df901dd1a020c38d1c40253997e517)

I can say I remember this being intentional to change the behavior that if the user did not specify a condition for a node, it would run only when its parents have run at least once since the last time it ran. This may not necessarily be in the same PASS, though. Here's an example where the current and proposed changes differ:

```a = pnl.TransferMechanism(name='a')
b = pnl.TransferMechanism(name='b')
c = pnl.TransferMechanism(name='c')
d = pnl.TransferMechanism(name='d')

comp = pnl.Composition(pathways=[[a, c, d], [a, b, d]])

comp.scheduler.add_condition_set({
    b: pnl.EveryNCalls(a, 2),
    c: pnl.EveryNCalls(a, 3),
})

comp.run(inputs={a: 1})
print(comp.scheduler.execution_list)
```

Current:
```
{'Composition-1': [{(TransferMechanism a)}, {(TransferMechanism a)}, {(TransferMechanism b)}, {(TransferMechanism a)}, {(TransferMechanism c)}, {(TransferMechanism d)}]}
```

Proposed:
```
{'Composition-1': [{(TransferMechanism a)}, {(TransferMechanism a)}, {(TransferMechanism b)}, {(TransferMechanism a)}, {(TransferMechanism c)}, {(TransferMechanism a)}, {(TransferMechanism b)}, {(TransferMechanism a)}, {(TransferMechanism a)}, {(TransferMechanism c), (TransferMechanism b)}, {(TransferMechanism d)}]}
```

I'm not sure how to rewrite it without using an EveryNCalls condition. Maybe @jdcpni can weigh in to confirm the best default behavior in this case?

jvesely:
The code doesn't consider parents though. It considers nodes in the previous consideration slot.
in the case of `a-b-c` and `a-c`, a->atpass(x), b->always. `c` would run after `b` irrespective of `a`.

In your example, you can add a-f. f->never, and it will prevent `d` from ever executing.
was the intention to have "all dependencies that are in earlier consideration slots"?

kmantel:
> The code doesn't consider parents though. It considers nodes in the previous consideration slot.
> in the case of `a-b-c` and `a-c`, a->atpass(x), b->always. `c` would run after `b` irrespective of `a`.
> 
> In your example, you can add a-f. f->never, and it will prevent `d` from ever executing.
> was the intention to have "all dependencies that are in earlier consideration slots"?

You're right. Based on the documentation changes when this was made, I think it's a bug. An even worse example is `a->b` and `c` independent. `b` would then depend on `c`. I think the following is correct, even in the case that you use a Never and some other nodes don't end up executing as a result.

```diff
diff --git a/psyneulink/core/scheduling/scheduler.py b/psyneulink/core/scheduling/scheduler.py
index bad120e01d2..689fc3b9ad5 100644
--- a/psyneulink/core/scheduling/scheduler.py
+++ b/psyneulink/core/scheduling/scheduler.py
@@ -573,7 +573,7 @@ class Scheduler(JSONDumpable):
                         break
 
                 if node_index > 0:
-                    dependencies = list(self.consideration_queue[i - 1])
+                    dependencies = list(self.dependency_dict[node])
                     if len(dependencies) == 1:
                         cond = EveryNCalls(dependencies[0], 1)
                     elif len(dependencies) > 1:
```

jvesely:
> > The code doesn't consider parents though. It considers nodes in the previous consideration slot.
> > in the case of `a-b-c` and `a-c`, a->atpass(x), b->always. `c` would run after `b` irrespective of `a`.
> > In your example, you can add a-f. f->never, and it will prevent `d` from ever executing.
> > was the intention to have "all dependencies that are in earlier consideration slots"?
> 
> You're right. Based on the documentation changes when this was made, I think it's a bug. An even worse example is `a->b` and `c` independent. `b` would then depend on `c`. I think the following is correct, even in the case that you use a Never and some other nodes don't end up executing as a result.
> 
> ```diff
> diff --git a/psyneulink/core/scheduling/scheduler.py b/psyneulink/core/scheduling/scheduler.py
> index bad120e01d2..689fc3b9ad5 100644
> --- a/psyneulink/core/scheduling/scheduler.py
> +++ b/psyneulink/core/scheduling/scheduler.py
> @@ -573,7 +573,7 @@ class Scheduler(JSONDumpable):
>                          break
>  
>                  if node_index > 0:
> -                    dependencies = list(self.consideration_queue[i - 1])
> +                    dependencies = list(self.dependency_dict[node])
>                      if len(dependencies) == 1:
>                          cond = EveryNCalls(dependencies[0], 1)
>                      elif len(dependencies) > 1:
> ```

thanks.
Either way, I think I'll need to restore support for `EveryNCalls`, the special case of N=1 should be doable with the existing tracking.
The other bug is that this code is entirely avoided in the standard compilation path.

PR #1914, opened 2021-02-17
fix(scheduler): default condition dependencies

kmantel:
condition incorrectly depended on the execution of all nodes in the previous consideration set, instead of its structural parents


buildbot-princeton:
Build finished.

PR #1916, opened 2021-02-18
llvm/scheduler: Restore and reimplement 'EveryNCalls' scheduling condition

jvesely:
We can match the semantics in N=1 special case.
This is needed to support autogenerated conditions for nodes that don't have any.

Fixes Python-PTX execution mode since dac19d4fb6df901dd1a020c38d1c40253997e517 ("llvm/scheduler: Remove 'EveryNCalls' scheduling condition")
Partially reverts dac19d4fb6df901dd1a020c38d1c40253997e517 ("llvm/scheduler: Remove 'EveryNCalls' scheduling condition")

### DIFF
--- a/psyneulink/core/llvm/helpers.py
+++ b/psyneulink/core/llvm/helpers.py
@@ -12,7 +12,7 @@ from llvmlite import ir
 from contextlib import contextmanager
 from ctypes import util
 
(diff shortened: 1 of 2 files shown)


jvesely:
retest this please

PR #1920, opened 2021-02-23
Parameter fitting of compositions using MLE via probability density approximation

davidt0x:
Initial implementation of parameter estimation with compositions using maximum likelihood estimation (MLE) via probability density approximation (PDA).


requirements.txt
```diff
@@ -13,4 +13,6 @@ toposort<1.7
 torch<1.8.0; sys_platform != 'win32' and platform_machine == 'x86_64' and platform_python_implementation == 'CPython'
 typecheck-decorator<=1.2
 leabra-psyneulink<=0.3.2
+pandas<=1.0.5
+psyneulink-fastkde>=1.0.19
```
jvesely, line 17: requirements have to have upper bound to prevent breakage on update

davidt0x:
Hold off on any merge please, I am still trying to track down a bug and another test for it.

PR #1928, opened 2021-02-28
tests: Add fixtures to control parametrization of execution_mode

jvesely:
Add `func_mode`, `mech_mode`, `comp_mode`, and `autodiff_mode` fixtures.
This replaces most of the manual mode parametrization.
It simplifies the code and moves control over the tested execution modes to a single location.
There are 23 locations that still use custom mode parametrization;
some tests (10) that use `composition.execute` instead of `run` and therefore omit `*Run` modes,
or scheduler tests that omit `LLVM` mode.

Several tests were extended to include PTX variants and use the new `mode` parametrization.
Three duplicate tests testing generator inputs were dropped.


jvesely:
This change is a prerequisite to changing `execution_mode` from strings to `enum.Flags`.
Without this consolidation, mode parametrization needs to be changed in hundreds of places,
With this change, it's down to ~25.

SamKG (approved):

kmantel (approved):

PR #1933, opened 2021-03-03
Fix jupyter detection for rich while running on colab

davidt0x:
Rich does not detect it is running in a jupyter notebook on Google colab. This is a bug in rich for which I have made a pull request.
See [here](https://github.com/willmcgugan/rich/pull/1085). Until this is accepted I suggest we just monkey patch rich.console._is_jupyter. If the pull request is accepted then this can be reverted.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,4 +13,4 @@ toposort<1.7
 torch<1.8.0; sys_platform != 'win32' and platform_machine == 'x86_64' and platform_python_implementation == 'CPython'
 typecheck-decorator<=1.2
 leabra-psyneulink<=0.3.2
-rich<=9.11.0
+rich>=9.11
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -13,4 +13,4 @@ toposort<1.7
 torch<1.8.0; sys_platform != 'win32' and platform_machine == 'x86_64' and platform_python_implementation == 'CPython'
 typecheck-decorator<=1.2
 leabra-psyneulink<=0.3.2
-rich<=9.11.0
+rich>=9.11
```
jvesely, line 16: this should preserve the upper bound even if a new lower bound needs to be added

davidt0x: Sounds good to me.

PR #1939, opened 2021-03-08
llvm: Implement multihreaded parallel evaluate

jvesely:
Generate samples locally instead of preparing the entire array in Python.
Add evaluate range wrapper.
Split the total number of evaluations evenly across all cores.


jvesely:
all cuda tests pass on my gf905m

SamKG (approved):

PR #1944, opened 2021-03-09
Refactor/pure functions

SamKG:
@kmantel 
@jvesely 
this is a draft of the pure transferfunction refactor - I could definitely use some input on if this is a good approach, or any other suggestions.

autocompiled udf code isn't in this draft yet - i want to completely nail down pure functions first

### DIFF
--- a/psyneulink/core/components/functions/transferfunctions.py
+++ b/psyneulink/core/components/functions/transferfunctions.py
@@ -130,7 +130,40 @@ class TransferFunction(Function_Base):
 
         return builder
 
+    @handle_external_context()
+    def function(self,
+                 variable=None,
+                 context=None,
+                 params=None,
+                 target_set=None,
+                 derivative=False,
+                 **kwargs):
+        # Validate variable and assign to variable, and validate params
+        variable = self._check_args(variable=variable,
+                                    context=context,
+                                    params=params,
+                                    target_set=target_set,
+                                    )
+        def _get_param(p):
+            try:
+                return self._get_current_parameter_value(p.name, context)
+            except:
+                print(e)
+        modulable_params = {p.name: _get_param(p) for p in self.parameters}
+        if params is None:
+            params = {}
+        parameters = {**modulable_params, **params}
+        parameters["variable"] = variable
+        if derivative:
+            return self._derivative(**kwargs, **parameters)
+        else:
+            value = self._function(**kwargs, **parameters)
+            self.most_recent_context = context
+            self.parameters.value._set(value, context=context)
+            return self.convert_output_type(value)
 
+    def derivative(self, *args, **kwargs):
+        return self.function(derivative=True, *args, **kwargs)
 # **********************************************************************************************************************
 #                                                 Identity
 # **********************************************************************************************************************
@@ -401,11 +434,11 @@ class Linear(TransferFunction):  # ---------------------------------------------
 
         builder.store(val, ptro)
 
-    def _function(self,
-                 variable=None,
-                 context=None,
-                 params=None,
-                 ):
+    @staticmethod
+    def _function(variable=None,
+                  slope=None,
+                  intercept=None,
+                  **_):
         """
 
         Arguments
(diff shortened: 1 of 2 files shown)


psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -130,7 +130,40 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, arg_in, arg_out,
...
+                                    )
+        def _get_param(p):
+            try:
+                return self._get_current_parameter_value(p.name, context)
+            except:
+                print(e)
+        modulable_params = {p.name: _get_param(p) for p in self.parameters}
+        if params is None:
+            params = {}
+        parameters = {**modulable_params, **params}
```
kmantel, line 155:
```suggestion
        parameters = {**modulable_params, **kwargs, **params}
```

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -130,7 +130,40 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, arg_in, arg_out,
...
+        parameters["variable"] = variable
+        if derivative:
+            return self._derivative(**kwargs, **parameters)
+        else:
+            value = self._function(**kwargs, **parameters)
```
kmantel, line 160:
```suggestion
            return self._derivative(**parameters)
        else:
            value = self._function(**parameters)
```

psyneulink/core/components/functions/transferfunctions.py
kmantel: Adding kwargs here prevents multiple keyword arguments error on ex: `pnl.Linear()(variable=5, slope=10, intercept=1)`

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -1559,9 +1543,6 @@ def derivative(self, input, output=None, context=None):
         derivative :  number or array
 
         """
-        gain = self._get_current_parameter_value(GAIN, context)
-        leak = self._get_current_parameter_value(LEAK, context)
-
         input = np.asarray(input).copy()
         input[input>0] = gain
         input[input<=0] = gain * leak
```
kmantel, line 1548:
```suggestion
        variable = np.asarray(variable).copy()
        variable[variable>0] = gain
        variable[variable<=0] = gain * leak
```

also line 1550
```
 	        return variable
```

kmantel (changes requested):

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -425,55 +458,13 @@ def _function(self,
...
-        Slope of function :  number or array
-
-        """
-
-        return self._get_current_parameter_value(SLOPE, context)
+    @staticmethod
+    def _derivative(variable=None,
+                    slope=None,
+                    **_):
+        return np.reshape(slope, variable.shape)
```
jvesely, line 467: this should use `broadcast` instead of `reshape`. reshape needs compatible shapes (same number of elements)

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -401,11 +434,11 @@ def _gen_llvm_transfer(self, builder, index, ctx, vi, vo, params, state, *, tags
...
-    def _function(self,
-                 variable=None,
-                 context=None,
-                 params=None,
-                 ):
+    @staticmethod
+    def _function(variable=None,
+                  slope=None,
+                  intercept=None,
+                  **_):
```
jvesely, line 441: This should update the argument documentation

jvesely, line 441: Why do you need hidden kwargs?

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -130,7 +130,40 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, arg_in, arg_out,
...
+        def _get_param(p):
+            try:
+                return self._get_current_parameter_value(p.name, context)
+            except:
+                print(e)
+        modulable_params = {p.name: _get_param(p) for p in self.parameters}
+        if params is None:
+            params = {}
+        parameters = {**modulable_params, **params}
+        parameters["variable"] = variable
```
jvesely, line 156: shouldn't this update the `variable` param in the current context?

psyneulink/core/components/functions/transferfunctions.py
jvesely, line 160: This is a rather ugly extra redirection. why can't you just move the checks to a helper method and call that from both `function` and `derivative`?

psyneulink/core/components/functions/transferfunctions.py
```diff
@@ -130,7 +130,40 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, arg_in, arg_out,
...
+        if derivative:
+            return self._derivative(**kwargs, **parameters)
+        else:
+            value = self._function(**kwargs, **parameters)
+            self.most_recent_context = context
+            self.parameters.value._set(value, context=context)
+            return self.convert_output_type(value)
 
+    def derivative(self, *args, **kwargs):
+        return self.function(derivative=True, *args, **kwargs)
```
jvesely, line 166: This should be:
```
params = self._process_params(*args, **kwargs)
return self._derivative(params)
```
I shoudl also allow you to drop the `**_` ugliness below.

jvesely:
Other than the few design issues. This is missign two major design justifications:

1.) Can all `Functions` be converted? What about the likes of `DDI`, `ContentAddressableMemory`?
If the compiler needs to be improved to handle those, converting TransferFunctions just removes cases that are easily supported and can be targeted as intermediate steps.

2.) How is the new design going to be enforced? Is every new Function expected to conform to the new format? If so what kind of help/guidance/enforcement will there be to guarantee that? Will there be a need for similar cleanup every time few more Functions are added?

PR #1945, opened 2021-03-10
Reenable reportOutputPref 'params' option

kmantel:
- add tests to ensure `'params'` option does not break again
- add workaround to Preference validation to mimic pre- 8c0eecc2c674a481138420f703ee661d39239052 behavior (4460fe5ea499441d8f0a255b62c4840de9c2b568)


jdcpni (commented):
Looks good.

PR #1957, opened 2021-03-19
bug(ControlMechanism): ensure simulation data and execution_id is unique

kmantel:
Make generated simulation execution_ids more verbose and unique, ensuring no
simulations reuse state.

If a Composition and one of its nested Compositions both run simulations,
execution ids may be inadvertently reused. See test_input_type_equivalence for
an example.

By default the outer Composition 'ocomp' runs with execution_id 'ocomp'. When it
simulates first, the generated execution_id will be 'ocomp-sim-0'. The inner
Composition 'icomp' will run during this simulation. Later, during
non-simulation running of ocomp, icomp will run under execution_id 'ocomp' and
simulate itself, also generating 'ocomp-sim-0'. Note that while in this case
there is no control allocation is directly specified for the simulation (hence
it not being included in the simulation execution_id), it can also happen when
the same allocation is specified in both situations.

The reuse of simulation execution_ids may cause reuse of state when not deleted.

### DIFF
--- a/psyneulink/core/components/mechanisms/modulatory/control/controlmechanism.py
+++ b/psyneulink/core/components/mechanisms/modulatory/control/controlmechanism.py
@@ -566,6 +566,7 @@ import itertools
 import numpy as np
 import threading
 import typecheck as tc
+import uuid
 import warnings
 
 from psyneulink.core import llvm as pnlvm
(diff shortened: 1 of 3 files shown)


kmantel:
retest this please

Issue #1956, opened 2021-03-18
Performance issues for large AutodiffCompositions


tylergiallanza:
I'm encountering performance issues in the first call to the `learn` function of an AutodiffComposition. The performance of the first call to the `learn` method seems to be highly dependent on the size of the mechanisms in the composition, scaling super-linearly. Consider this code snippet, adapted from the documentation:

```  
    print(f's1={s1},s2={s2}')
    start_t = time.time()
    my_mech_1 = pnl.TransferMechanism(function=pnl.Linear, size = s1)
    my_mech_2 = pnl.TransferMechanism(function=pnl.Linear, size = s2)
    my_projection = pnl.MappingProjection(matrix=np.random.randn(s1,s2),
                        sender=my_mech_1,
                        receiver=my_mech_2)
    # Create AutodiffComposition
    my_autodiff = pnl.AutodiffComposition()
    my_autodiff.add_node(my_mech_1)
    my_autodiff.add_node(my_mech_2)
    my_autodiff.add_projection(sender=my_mech_1, projection=my_projection, receiver=my_mech_2)
    print(f'  Time to create: {time.time()-start_t}')
    in_patterns = np.random.binomial(1,.1,size=(10,s1))
    out_patterns = np.random.binomial(1,.2,size=(10,s2))
    start_t = time.time()
    k = my_autodiff.learn(inputs = {'inputs':{my_mech_1:in_patterns}, 
                             'targets':{my_mech_2:out_patterns}})
    print(f'  First run\'s time: {time.time()-start_t}')
    start_t = time.time()
    k = my_autodiff.learn(inputs = {'inputs':{my_mech_1:in_patterns}, 
                             'targets':{my_mech_2:out_patterns}})
    print(f'  Second run\'s time: {time.time()-start_t}')

```

I ran this code for different values of `s1` and `s2`, yielding this:

| s1  | s1 | time to create composition | time for first call to learn | time for second call to learn |
| ------------- | ------------- | ------------- |  ------------- |  ------------- |
| 1 | 2  | .5 sec | 1 sec | .03 sec |
| 10 | 20  | .5 sec | 1 sec | .03 sec |
| 100 | 200  | .8 sec | 8 sec | .03 sec |
| 200 | 400  | 1.8 sec | 23 sec | .03 sec |
| 200 | 800  | 3.4 sec | 85 sec | .03 sec |

[...]

Issue #1955, opened 2021-03-18
Speed issue when adding linear processing pathway


tylergiallanza:
It appears that the `add_linear_processing_pathway` method for Compositions is inefficient for large mechanisms. Consider the following code sample:

```
comp = pnl.Composition()
mech1 = pnl.TransferMechanism(default_variable=np.zeros(100))
mech2 = pnl.TransferMechanism(default_variable=np.zeros(1000))
comp.add_linear_processing_pathway([mech1,mech2])
```

This takes 32.3 seconds to run on my machine (using the `%%timeit` magic in Jupyter notebook). Creating the pathway manually, however:
```
comp = pnl.Composition()
mech1 = pnl.TransferMechanism(default_variable=np.zeros(100))
mech2 = pnl.TransferMechanism(default_variable=np.zeros(1000))
comp.add_node(mech1)
comp.add_node(mech2)
map_ = pnl.MappingProjection(matrix=np.random.rand(100, 1000),
                               name="map_h12"
                            )
comp.add_projection(projection=map_,sender=mech1,receiver=mech2)
```
Takes only 3 seconds to run.

P.S. the time to run the `add_linear_processing_pathway` seems to scale super-linearly w.r.t the size of `mech2` in the above example:
| # of Units in Mech2  | Time to create composition |
| ------------- | ------------- |
| 100  | 1 second  |
| 1000  | 30 seconds  |
| 2000  | 90 seconds  |

jdcpni:
FE(everyone’s)I:  its in the call to analyze_graph (25 of the 30 seconds).   More to explore.


jdc



On Mar 18, 2021, at 5:22 PM, tylergiallanza ***@***.******@***.***>> wrote:



It appears that the add_linear_processing_pathway method for Compositions is inefficient for large mechanisms. Consider the following code sample:

comp = pnl.Composition()
mech1 = pnl.TransferMechanism(default_variable=np.zeros(100))
mech2 = pnl.TransferMechanism(default_variable=np.zeros(1000))
comp.add_linear_processing_pathway([mech1,mech2])


This takes 32.3 seconds to run on my machine (using the %%timeit magic in Jupyter notebook). Creating the pathway manually, however:

comp = pnl.Composition()
mech1 = pnl.TransferMechanism(default_variable=np.zeros(100))
mech2 = pnl.TransferMechanism(default_variable=np.zeros(1000))
comp.add_node(mech1)
comp.add_node(mech2)
[...]

PR #1963, opened 2021-03-20
Component: speed up noise generation for common cases

kmantel:
When noise is numeric (not generated at Component run time by functions), do not
apply it element-by-element, which can get very slow for large variables.

### DIFF
--- a/psyneulink/core/components/component.py
+++ b/psyneulink/core/components/component.py
@@ -3199,6 +3199,10 @@ class Component(JSONDumpable, metaclass=ComponentsMeta):
 
         var = convert_all_elements_to_np_array(copy.deepcopy(var), cast_from=int, cast_to=float)
 
+        # ignore zero-length variables (e.g. empty Buffer)
+        if var.shape == (0, ):
+            return var
+
         # handle simple wrapping of a Component (e.g. from ParameterPort in
         # case of set after Component instantiation)
         if (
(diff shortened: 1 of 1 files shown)


kmantel:
This resolves the speed problems in #1955 and #1956, which are good examples of the common case mentioned above

psyneulink/core/components/component.py
```diff
@@ -3223,8 +3227,12 @@ def fill_recursively(arr, value, indices=()):
         except AttributeError:
             pass
 
-        fill_recursively(var, param)
-        return var
+        param_arr = np.asarray(param)
+        if param_arr.dtype == int or param_arr.dtype == float:
```
jvesely, line 3231: can this use `np.isscalar` ?

psyneulink/core/components/component.py
```diff
@@ -3196,6 +3196,10 @@ def fill_recursively(arr, value, indices=()):
 
         var = convert_all_elements_to_np_array(copy.deepcopy(var), cast_from=int, cast_to=float)
 
+        # occurs for some Buffer objects
```
jvesely, line 3199: It'd be more useful to say why it is ok that it occurs and it needs special handling (rather than fix on the Buffer side).

psyneulink/core/components/component.py
kmantel: Not quite, np.isscalar returns False for 0-dim numeric arrays (and longer), though in many cases they can be reshaped to match variable.

kmantel:
retest this please

psyneulink/core/components/component.py
jvesely: I meant `np.isscalar(param_arr.dtype)`, that should cover both `int` and `float`

kmantel: Oh, I see. The main point is to check for a fully regular, numeric array. I suppose `np.unsignedinteger` would be a valid dtype though I don't think we use it anywhere. In this case, maybe it makes sense to just rewrite as `if param_arr.dtype != object`, because any other invalid types for param/noise should be instead caught at initial assignment and would fail later regardless. I have already used that for a similar check elsewhere

Issue #1967, opened 2021-03-23
Optional dependency on PyTorch is not clear


SamKG:
Currently, PyTorch is an optional dependency for PsyNeuLink.
A user who does not have pytorch installed, but wants to use an AutodiffComposition, may try to index an AutodiffComposition with:
```
import psyneulink as pnl
comp = pnl.AutodiffComposition()
```

This code currently throws an opaque error; It would be better to warn the user when they try to index AutodiffComposition that pytorch is not installed.

PR #914, opened 2018-09-24
Merge autodiff composition into devel

karoara:

PR #1968, opened 2021-03-23
autodiffcomp: Error when AutodiffComposition is indexed without PyTor…

SamKG:
…ch (#1967)


kmantel:
There are some checks already ([autodiffcomposition](https://github.com/PrincetonUniversity/PsyNeuLink/blob/d105adda0adad27d70d32f8f8d6ab8bd5f79086d/psyneulink/library/compositions/autodiffcomposition.py#L235-L237) and [pytorchmodelcreator](https://github.com/PrincetonUniversity/PsyNeuLink/blob/d105adda0adad27d70d32f8f8d6ab8bd5f79086d/psyneulink/library/compositions/pytorchmodelcreator.py#L26-L28])), are those missing cases?

SamKG:
> 
> 
> There are some checks already ([autodiffcomposition](https://github.com/PrincetonUniversity/PsyNeuLink/blob/d105adda0adad27d70d32f8f8d6ab8bd5f79086d/psyneulink/library/compositions/autodiffcomposition.py#L235-L237) and [pytorchmodelcreator](https://github.com/PrincetonUniversity/PsyNeuLink/blob/d105adda0adad27d70d32f8f8d6ab8bd5f79086d/psyneulink/library/compositions/pytorchmodelcreator.py#L26-L28%5D)), are those missing cases?

Actually, those cases are never hit; Since AutodiffCompositions are not imported when torch is not found, pnl.AutodiffComposition does not exist to begin with, thus never running the existing checks.

kmantel:
I see, what about something like 

```diff
diff --git a/psyneulink/library/compositions/__init__.py b/psyneulink/library/compositions/__init__.py
index 1f5039502e..422ebf9ebf 100644
--- a/psyneulink/library/compositions/__init__.py
+++ b/psyneulink/library/compositions/__init__.py
@@ -3,11 +3,6 @@ from .compositionrunner import *
 __all__ = list(regressioncfa.__all__)
 __all__.extend(compositionrunner.__all__)
 
-try:
-    import torch
+from .autodiffcomposition import *
 
-    from .autodiffcomposition import *
-
-    __all__.extend(autodiffcomposition.__all__)
-except ImportError:
-    pass
+__all__.extend(autodiffcomposition.__all__)
```

instead, which uses the error already in `AutoDiffComposition` with the benefit that the actual class is imported? It looks like the original import check is from #914 and the error message and (seemingly redundant) check in compositions/__init__.py in #923.

SamKG:
> 
> 
> I see, what about something like
> 
> ```diff
> diff --git a/psyneulink/library/compositions/__init__.py b/psyneulink/library/compositions/__init__.py
> index 1f5039502e..422ebf9ebf 100644
> --- a/psyneulink/library/compositions/__init__.py
> +++ b/psyneulink/library/compositions/__init__.py
> @@ -3,11 +3,6 @@ from .compositionrunner import *
>  __all__ = list(regressioncfa.__all__)
>  __all__.extend(compositionrunner.__all__)
>  
> -try:
> -    import torch
> +from .autodiffcomposition import *
>  
> -    from .autodiffcomposition import *
> -
> -    __all__.extend(autodiffcomposition.__all__)
> -except ImportError:
> -    pass
> +__all__.extend(autodiffcomposition.__all__)
> ```
> 
> instead, which uses the error already in `AutoDiffComposition` with the benefit that the actual class is imported? It looks like the original import check is from #914 and the error message and (seemingly redundant) check in compositions/**init**.py in #923.

Fixed!
Thanks for the suggestion!

PR #1981, opened 2021-04-02
fix(PytorchModelCreator): use new context to init .execution_sets

kmantel:
in \_\_init\_\_, a full trial was run in the passed context, potentially messing up
future counts or clocks in that context


kmantel:
retest this please

PR #1998, opened 2021-04-10
json/mdf: ignore learning components when regenerating script

kmantel:
it was not reliable/tested in json examples, and it's unclear if there are any
actual examples of manual learning pathways for psyneulink. (all learning
appears to be done via premade methods like
Composition.add_backpropagation_learning_pathway)


kmantel:
retest this please

PR #2001, opened 2021-04-13
fix(OptimizationControlMechanism): don't immediately create shadowed projection for features

kmantel:
Fixes a problem in which an incorrect shadowed projection is created from an
OCM/controller to a node. Specifically, consider a composition O with a nested
composition I, containing a node A. If the controller of O has A as its feature,
the controller should only control A and receive shadowed input through the
parameter CIM of I. Prior to this commit, the controller of O received a
shadowed input projection directly from A, created during
_instantiate_input_ports during construction of the controller. This step is
unnecessary and incorrect as shadowed input projections are created during the
call to Composition.add_controller.


kmantel:
Visualization showing the extra shadowed projection of A in I 

old


vs. with only the correct shadowed projection in O

new

PR #2022, opened 2021-04-23
requirements: bump pytorch to 1.8.x

kmantel:
1.8.0 required in MDF (https://github.com/ModECI/MDF/pull/50) which will soon require psyneulink (https://github.com/ModECI/MDF/pull/44)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ networkx<2.6
 numpy<1.20.3
 pillow<8.1.0
 toposort<1.7
-torch<1.8.0; sys_platform != 'win32' and platform_machine == 'x86_64' and platform_python_implementation == 'CPython'
+torch>=1.8.0, <1.9.0; sys_platform != 'win32' and platform_machine == 'x86_64' and platform_python_implementation == 'CPython'
 typecheck-decorator<=1.2
 leabra-psyneulink<=0.3.2
 rich>=10.1, <10.2


jvesely:
> 1.8.0 required in MDF ([ModECI/MDF#50](https://github.com/ModECI/MDF/pull/50)) which will soon require psyneulink ([ModECI/MDF#44](https://github.com/ModECI/MDF/pull/44))

We should include minimum requirements in the dep: `torch >=1.8.1, <1.9.0`

SamKG (approved):

jvesely:
We can also probably drop some of the selectors. `pip install torch` should work on 64bit Python on windows/mac/linux/linux-aarch64.
On the other hand there is no selector to detect 32bit python installation on 64bit systems, so that will need to be handled via `setup.py`.
I've been trying to clean it up, but got stuck on not finding a good way to have a default-on optional feature.

kmantel:
I haven't seen anything like a default-on optional requirement, maybe we should move it to a dev requirement? Either way, there isn't a need to rush this right now for MDF if it should wait until we figure out what to do with the requirement in general.

jvesely:
> I haven't seen anything like a default-on optional requirement, maybe we should move it to a dev requirement? Either way, there isn't a need to rush this right now for MDF if it should wait until we figure out what to do with the requirement in general.

I think this can be bumped independently of other changes to how pytorch is handled/cleaned up later

PR #2034, opened 2021-05-02
github-actions: Consolidate PNL package setup

jvesely:
Move dependency handling to common action.
Add wheel cache to doc-compare workflow.
Use multiple cores to generate docs.
Allow direct dependency on torch on windows.


jvesely:
Ignoring docs compare. It won't run because of the changes in CI scripts.
Ignoring lgtm.com. It's stuck on fetching commits and there are not python changes anyway.

PR #2036, opened 2021-05-04
learning: Writeback params at end of learning instead of at each batch

SamKG:
This results in a ~5% speedup for a smaller model (and should scale for larger models as well)

### DIFF
--- a/psyneulink/library/compositions/autodiffcomposition.py
+++ b/psyneulink/library/compositions/autodiffcomposition.py
@@ -403,6 +403,8 @@ class AutodiffComposition(Composition):
         self.parameters.tracked_loss_count._set(0, context=context, skip_history=True, skip_log=True)
         optimizer.step()
         self.parameters.pytorch_representation._get(context=context).detach_all()
+
+    def _writeback_parameters(self, context):
         self.parameters.pytorch_representation._get(context).copy_weights_to_psyneulink(context)
 
     def _gen_llvm_function(self, *, ctx:pnlvm.LLVMBuilderContext, tags:frozenset):
--- a/psyneulink/library/compositions/autodiffcomposition.py
+++ b/psyneulink/library/compositions/autodiffcomposition.py
@@ -128,6 +128,7 @@ Class Reference
 
 """
 import logging
+from typing import Union
 
 import numpy as np
 
@@ -248,7 +249,7 @@ class AutodiffComposition(Composition):
         self.optimizer_type = optimizer_type
         self.loss_spec = loss_spec
         self.refresh_losses = refresh_losses
-        self._built_pathways = False
+        self._built_backprop_pathway = False
         self.weight_decay = weight_decay
         self.force_no_retain_graph = force_no_retain_graph
         self.loss = None
@@ -270,13 +271,14 @@ class AutodiffComposition(Composition):
 
     # CLEANUP: move some of what's done in the methods below to a "validate_params" type of method
     @handle_external_context()
-    def _build_pytorch_representation(self, context=None):
+    def _build_pytorch_representation(self, context=None, scheduler=None):
         if self.scheduler is None:
             self.scheduler = Scheduler(graph=self.graph_processing)
         if self.parameters.pytorch_representation._get(context=context) is None:
             model = PytorchModelCreator(composition=self,
                                         device=self.device,
-                                        context=context)
+                                        context=context,
+                                        passed_scheduler=scheduler)
 
             self.parameters.pytorch_representation._set(model, context, skip_history=True, skip_log=True)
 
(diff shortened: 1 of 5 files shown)


psyneulink/library/compositions/compositionrunner.py
```diff
@@ -228,6 +228,10 @@ def run_learning(self,
         # return results from last epoch
         results = self._composition.parameters.results.get(context)[-1 * num_epoch_results:]
 
+        # FIXME: This currently does not work for pnlvm.ExecutionMode.Auto, since we have no way of knowing what mode was taken
+        if hasattr(self._composition, "_writeback_parameters") and (execution_mode == pnlvm.ExecutionMode.Python):
```
jvesely, line 232: can this ever run anything else than Autodiff composition?

SamKG: Yes; PsyNeuLink has learning mechanisms (https://princetonuniversity.github.io/PsyNeuLink/LearningMechanism.html) that are able to be learned by calling `composition.learn`. We only need the writeback in the case of autodiff though, since we have a pytorch backend that maintains separate copies of the weights

jvesely: so `self._composition` can be a learning mechanism?

SamKG: it can be a non-autodiff compostion, with learning mechanisms inside of it

jvesely: what if there is another autodiff nested inside? can you call learn on the outer composition?

SamKG: ah, I see your point.
Yes, this is definitely an edge-case that was missed. Good catch!

buildbot-princeton:

psyneulink/library/compositions/compositionrunner.py
SamKG: Fixed!

psyneulink/library/compositions/autodiffcomposition.py
```diff
@@ -528,16 +556,20 @@ def execute(self,
...
+                    report_num=report_num,
+                    scheduler=scheduler,
+                    content='trial_end',
+                    context=context)
+
+            # FIXME: In nested autodiff, the scheduler is unable to correctly update the timescale
+            try:
+                scheduler.get_clock(context)._increment_time(TimeScale.TRIAL)
+            except KeyError as e:
+                pass
```
kmantel, line 572:
```suggestion
            scheduler.get_clock(context)._increment_time(TimeScale.TRIAL)
```
I think this would throw KeyError when run outside of nesting, right? In which case I think we want to have a clock for context and run this line every time:

```diff
diff --git a/psyneulink/library/compositions/autodiffcomposition.py b/psyneulink/library/compositions/autodiffcomposition.py
index fc5c948c82..1cf0ada9a3 100644
--- a/psyneulink/library/compositions/autodiffcomposition.py
+++ b/psyneulink/library/compositions/autodiffcomposition.py
@@ -518,6 +518,11 @@ class AutodiffComposition(Composition):
         if scheduler is None:
             scheduler = self.scheduler
 
+        try:
+            scheduler.get_clock(context)
+        except KeyError:
+            scheduler._init_counts(context.execution_id, self.base_execution_id)
+
         if self._is_learning(context):
             # TBI: How are we supposed to use base_context and statefulness here?
             # TBI: can we call _build_pytorch_representation in _analyze_graph so that pytorch
```

kmantel (changes requested):
The uncommented changes look ok

psyneulink/library/compositions/autodiffcomposition.py
kmantel: I don't think it needs to be avoided because autodiff doesn't call Scheduler.run like other compositions, but I think it's better to do it as in the diff and suggestion. Otherwise, in the case the `get_clock(context)` call fails, there just won't be a time for context, rather than a time accurately reflecting the number of trials that were run.

tests/composition/test_autodiffcomposition.py
```diff
@@ -2449,6 +2449,67 @@ def test_xor_nested_train_then_no_train(self, num_epochs, learning_rate,
...
+        parentComposition.add_node(xor_autodiff_inner)
+        input_dict = {'inputs': {xor_autodiff_inner_input: xor_inputs }, 'targets': {xor_autodiff_inner_output: xor_targets}}
+        parent_input_dict = {
+            xor_autodiff_inner: input_dict,
+        }
+        result2 = parentComposition.learn(inputs=parent_input_dict, execution_mode=autodiff_mode, epochs=num_epochs, patience=patience, min_delta=min_delta)
+
+
+        result1 = np.array(result1).flatten()
+        assert np.allclose(np.array(result1).flatten(), np.array(result2).flatten())
```
jvesely, line 2511: drop `flatten`. the shapes should be known and included in the test.

jvesely:
Compiled and Python Parameters are now synced via shared memory.
Explicit synchronization is no longer needed.

PR #2044, opened 2021-05-10
jenkins: Include sources in pytest search scope

jvesely:
Test runs should include codestyle/docstyle checks on source files.

### DIFF
--- a/bin/jenkins_build.sh
+++ b/bin/jenkins_build.sh
@@ -16,9 +16,6 @@ if [ -z $env_name ]; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jvesely:
> I've just checked and it looks like jenkins_build.sh isn't what's actually used on jenkins. It looks like the current pytest command does include the psyneulink directory - have you noticed a case it's missing? Either way, I'll update the script to what's currently running, make jenkins use it, and then set up a new PR

thanks for checking. it was just a cleanup while working on github-actions. We can just drop the script if it's not used.

PR #2045, opened 2021-05-11
jenkins: copy actual build script used, to be target

kmantel:
`jenkins_build.sh` contains a slightly modified version of the script jenkins actually uses for the psyneulink_test_py* jobs. This can be called instead, so any changes to the script or the tox config can be tracked. This is working on the pnl36-copy job, that can be removed when this PR is merged or closed. 

My only concern is about tests automatically running on jenkins on a PR without admin approval. It looks like this is enabled by default for the [pull request builder plugin](https://plugins.jenkins.io/ghprb/), and it doesn't look overridden to me, but it would be good for someone else to double check. 

This is a followup to #2044

### DIFF
--- a/bin/jenkins_build.sh
+++ b/bin/jenkins_build.sh
@@ -1,104 +1,27 @@
-#!/bin/bash
-# you must run this script in the top level dir
-source ~/.bashrc
+#!/bin/bash -l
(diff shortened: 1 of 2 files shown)


jvesely:
Now that Jenkins has been retired, we can close this one.

PR #2048, opened 2021-05-11
github-actions: Prepare pre-release on tag on master

jvesely:
Two things happen when a tag is pushed to master:
 * the tagged commit is merged back to devel
 * a new pre-release is created (which trigger test-release workflow)

Creating a release needs PAT with public_repo privilege level in order
to trigger the test-release workflow.
The default actions token does not trigger workflows on repo actions.
If token is set to 'no-token' the pre-release creation is skipped.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,73 @@
+name: Prepare PNL release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Checkout sources
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+
+    - name: Check if on master
+      id: on_master
+      shell: bash
+      run: |
+        git fetch origin master --depth=100
+        git describe --always --tags
+        export ON_MASTER=$(git branch -a --contains ${{ github.ref }} | grep -q '^  remotes/origin/master$' && echo "master" || echo "")
+        echo "Found out: ${ON_MASTER}"
+        echo ::set-output name=on_master::$ON_MASTER
+
+    - name: Check for existing release with the reference tag
+      uses: actions/github-script@v4
+      id: exist_check
+      with:
+        script: |
+          tag = context.ref.split('/').pop()
+          console.log('running on:' + context.ref);
+          console.log('Looking for release for tag:' + tag);
+          try {
+            release_if_exists = await github.repos.getReleaseByTag({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag: tag
+            });
+            console.log('Release found at: ' + release_if_exists.data.html_url);
+            core.setOutput('exists', 'yes')
+          } catch (err) {
+            if (err.status == 404) {
+              console.log('Release not found.');
+              core.setOutput('exists', 'no')
+            } else {
+              throw err;
+            }
+          }
+
+    - name: Create Release
+      uses: actions/github-script@v4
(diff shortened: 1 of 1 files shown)


jvesely:
@kmantel , any concerns about this one?

kmantel (commented):
No objection in general, but I wonder if the "Merge release back to devel" step will work as intended? It looks to me that if the tag is pushed before required checks finish on master, this step will fail on the git push, stopping the rest of the action. I think something like [this](https://github.com/marketplace/actions/wait-for-check) for example would prevent that, although the tag-pusher checking this would work too.

jvesely:
> No objection in general, but I wonder if the "Merge release back to devel" step will work as intended? It looks to me that if the tag is pushed before required checks finish on master, this step will fail on the git push, stopping the rest of the action. I think something like [this](https://github.com/marketplace/actions/wait-for-check) for example would prevent that, although the tag-pusher checking this would work too.

Why would it fail on push? the tag is already present whether checks succeed or not.
There's a race if someone merges to devel in the time between checkout and merge, but that is unlikely.

kmantel:
> > No objection in general, but I wonder if the "Merge release back to devel" step will work as intended? It looks to me that if the tag is pushed before required checks finish on master, this step will fail on the git push, stopping the rest of the action. I think something like [this](https://github.com/marketplace/actions/wait-for-check) for example would prevent that, although the tag-pusher checking this would work too.
> 
> Why would it fail on push? the tag is already present whether checks succeed or not.
> There's a race if someone merges to devel in the time between checkout and merge, but that is unlikely.

Like this I think: when there is a merge to master, at least by PR, CI runs on master. If a tag pointing to master is pushed right after, this action will run, merging the tag into devel and pushing. But in this case, the CI on master/tag won't be finished, and since devel is a protected branch and the required jobs don't have successful status yet, the push will be rejected.

jvesely:
> > > No objection in general, but I wonder if the "Merge release back to devel" step will work as intended? It looks to me that if the tag is pushed before required checks finish on master, this step will fail on the git push, stopping the rest of the action. I think something like [this](https://github.com/marketplace/actions/wait-for-check) for example would prevent that, although the tag-pusher checking this would work too.
> > 
> > 
> > Why would it fail on push? the tag is already present whether checks succeed or not.
> > There's a race if someone merges to devel in the time between checkout and merge, but that is unlikely.
> 
> Like this I think: when there is a merge to master, at least by PR, CI runs on master. If a tag pointing to master is pushed right after, this action will run, merging the tag into devel and pushing. But in this case, the CI on master/tag won't be finished, and since devel is a protected branch and the required jobs don't have successful status yet for the proposed new devel, the push will be rejected.

right, branch protection. I'll it locally andupdate the PR. thanks

jvesely:
> > > No objection in general, but I wonder if the "Merge release back to devel" step will work as intended? It looks to me that if the tag is pushed before required checks finish on master, this step will fail on the git push, stopping the rest of the action. I think something like [this](https://github.com/marketplace/actions/wait-for-check) for example would prevent that, although the tag-pusher checking this would work too.
> > 
> > 
> > Why would it fail on push? the tag is already present whether checks succeed or not.
> > There's a race if someone merges to devel in the time between checkout and merge, but that is unlikely.
> 
> Like this I think: when there is a merge to master, at least by PR, CI runs on master. If a tag pointing to master is pushed right after, this action will run, merging the tag into devel and pushing. But in this case, the CI on master/tag won't be finished, and since devel is a protected branch and the required jobs don't have successful status yet for the proposed new devel, the push will be rejected.

I've dropped the 'merge back to devel' part. I think a separate workflow 'on check complete' is a better fit than trying to add waiting to this one.

jvesely:
@kmantel , any problems with the most recent version? This PR is holding back all other work on CI consolidation.

PR #2051, opened 2021-05-14
UserDefinedFunction: use signature instead of co_varnames for argument names

kmantel:


jvesely:
I think this could do with a few tests for those newly allowed functions.

PR #2057, opened 2021-05-24
llvm: Add generic conversion helper.

jvesely:
Use the helper to convert array indices and as_type.
Move float->int conversion from Index to Subscript.
Enable python3.9 CI jobs.


jvesely:
There are more changes coming, but this is enough to enable python 3.9

PR #2058, opened 2021-05-25
Drop support for Python 3.6

jvesely:
Drop python 3.6 CI jobs.
Bump python requirement.

### DIFF
--- a/.github/actions/install-pnl/action.yml
+++ b/.github/actions/install-pnl/action.yml
@@ -36,8 +36,7 @@ runs:
     - name: Python dependencies
       shell: bash
       run: |
-        # explicitly install numpy (https://github.com/pypa/pip/issues/9239)
-        python -m pip install --upgrade pip wheel $(grep numpy requirements.txt)
+        python -m pip install --upgrade pip wheel
         pip install -e .[${{ inputs.features }}]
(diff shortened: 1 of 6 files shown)


jvesely:
I've tested that the CI (including the release script works).

https://github.com/jvesely/PsyNeuLink/actions/runs/875922224

kmantel (approved):
One reason is numpy dropping support for 3.6, any others in particular? Anyway, I moved the required CI status to jenkins 3.7 so this can go through.

jvesely:
> One reason is numpy dropping support for 3.6, any others in particular? Anyway, I moved the required CI status to jenkins 3.7 so this can go through.

I think it was a more generic "support the last 3 versions of python" thing.
Python 3.6 EOL date is Dec 2021 and has been providing source-only releases for some time.

Linux and Windows can easily install a newer version from the package managers.
I don't really know how difficult it is to maintain up-to-date Python on mac.

This PR can wait until December, but I don't see any benefit at the cost of more CI jobs.

jvesely:
retest this please

jvesely:
looks like the pytest-3.6 job needs to be disabled entirely for this PR to go ahead

PR #2065, opened 2021-06-01
Add resettable seed Parameter

kmantel:
For Functions or classes that use `random_state`, add the `seed` Parameter that reseeds `random_state` on set, or at the first request of `random_state` per context when `seed` is modulated.

the following expected test outputs are changed because the `show_node_structure=True` argument to show_graph shows all ParameterPorts. Adding seed as a modulable parameter adds an extra html structure to these tests, and this is the only change:

```
tests/composition/test_show_graph.py::TestControl::test_no_nested_and_controler_name_with_space_in_it[{'show_controller': True, 'show_node_structure': True}]
tests/composition/test_show_graph.py::TestControl::test_nested_learning_test_with_user_specified_target_in_outer_composition[{'show_nested': False, 'show_cim': False, 'show_node_structure': True, 'show_learning': True}]
tests/composition/test_show_graph.py::TestControl::test_nested_learning_test_with_user_specified_target_in_outer_composition[{'show_nested': False, 'show_cim': True, 'show_node_structure': True, 'show_learning': True}]
tests/composition/test_show_graph.py::TestControl::test_nested_learning_test_with_user_specified_target_in_outer_composition[{'show_nested': 'nested', 'show_cim': False, 'show_node_structure': True, 'show_learning': True}]
tests/composition/test_show_graph.py::TestControl::test_nested_learning_test_with_user_specified_target_in_outer_composition[{'show_nested': 'nested', 'show_cim': True, 'show_node_structure': True, 'show_learning': True}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_nested': 'inset'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_nested': False}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_nested': 'nested'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_cim': True, 'show_nested': False}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_cim': True, 'show_nested': 'inset'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_cim': True, 'show_nested': 'nested'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_3_level_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_nested': False}]
tests/composition/test_show_graph.py::TestControl::test_of_show_3_level_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_nested': 'inset'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_3_level_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_nested': 'nested'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_3_level_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_cim': True, 'show_nested': False}]
tests/composition/test_show_graph.py::TestControl::test_of_show_3_level_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_cim': True, 'show_nested': 'inset'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_3_level_nested_show_cim_and_show_node_structure[{'show_node_structure': True, 'show_cim': True, 'show_nested': 'nested'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure_with_singleton_in_outer_comp_added_last[{'show_node_structure': True, 'show_nested': False}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure_with_singleton_in_outer_comp_added_last[{'show_node_structure': True, 'show_nested': 'inset'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure_with_singleton_in_outer_comp_added_last[{'show_node_structure': True, 'show_nested': 'nested'}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure_with_singleton_in_outer_comp_added_last[{'show_node_structure': True, 'show_cim': True, 'show_nested': False}]
tests/composition/test_show_graph.py::TestControl::test_of_show_nested_show_cim_and_show_node_structure_with_singleton_in_outer_comp_added_last[{'show_node_structure': True, 'show_cim': True, 'show_nested': 'inset'}]
```


psyneulink/core/components/functions/function.py
```diff
@@ -327,6 +327,18 @@ def _output_type_setter(value, owning_component):
     return value
 
 
+def _seed_setter(value, owning_component, context):
```
jvesely, line 330: does this work with param modulation? would a change in modulated value reseed the random state?

kmantel: I don't think so. As it is now, the value of a parameter is only replaced with its corresponding parameter port's value when that port exists. I'm thinking of some changes for additional actions when getting a potentially modulated value. Assuming no changes to whatever does the modulating, these changes would be based on how often the reseed should happen. Using the example of an OCM running a simulation and changing the seed for each, do we want a reseed at the first time the `random_state` is requested per simulation/context?

jvesely: I think that for simulation purposes it only needs to happen once. Something like delaying random state initialization until the first request should work (we should see a modulated seed at that point).
It'd still be nice if the (modulated) seed that was used for initialization was reflected in the seed param.

kmantel: OK, I'll assume only automatically reseeding once per context. 

>It'd still be nice if the (modulated) seed that was used for initialization was reflected in the seed param.

This is sort of at odds with an intentional (but now old) decision that the main parameter value tracked/stored/got would be the base value, not modulated. The modulated value would be accessible through `Component.get_mod_seed` (autogenerated) or `Component._get_current_parameter_value`. I think it might be worth splitting this to its own issue, but we could add something to get the modulated value like `Parameter._get_modulated` or an additional argument to `Parameter._get` for example. I think keeping `_get` for just the base value makes sense.

kmantel:
The latest changes should handle resetting random states when seed is modulated. (This commit has show_graph failures, which I'm sure is due to seed now being modulable and adding a new ParameterPort, but I'll double check this)

psyneulink/core/components/functions/function.py
jvesely: yeah, sounds reasonable. I think get_mod_seed does the job just fine.

PR #2067, opened 2021-06-05
tests: show_graph: parametrize for easier analysis on failure

kmantel:
When show_graph tests fail, it's difficult to determine which configurations fail and then to modify the expected output when the changes are ok. Using pytest.parametrize splits the tests by input to show_graph making this much easier


tests/composition/test_show_graph.py
```diff
@@ -233,28 +234,83 @@ def test_converging_pathways(self):
...
+        )
+    ]
 
+    @pytest.fixture(scope='session')
+    def _no_nested_and_controler_name_with_space_in_it_comp(self):
```
jvesely, line 255: This is a rather hacky missuse of fixtures. The function is called only in one place and there's no reason for it to be a fixture or a standalone function.
Making this a fixture moves the composition construction outside of the test with hard to read/diagnose error messaging if it fails.

kmantel: The main reason was that each parametrization runs the composition setup again, which caused the set of show_graph tests to run noticeably slower, at least locally. Using a scoped fixture avoids this, although depending on how much it affects GHA/jenkins (will check) it may not be worth worrying about. I admit it's ugly though.

tests/composition/test_show_graph.py
```diff
@@ -233,28 +234,83 @@ def test_converging_pathways(self):
         assert "out (1)" in b_label and "in (1)" in b_label
         assert "out (5)" in c_label and "in (5)" in c_label
 
-    def test_no_nested_and_controler_name_with_space_in_it(self):
+    # ------------------------------------------------------------------
+
+    _no_nested_and_controler_name_with_space_in_it_data = [
+        (
+            {'show_controller': False, 'output_fmt': 'source'},
```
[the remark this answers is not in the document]
kmantel: Oversight, will change. I don't expect any of these not to use 'source'

tests/composition/test_show_graph.py
kmantel: There's enough variance in the duration of the test suite from run to run that it's not clear how much time the fixtures save, so I've removed them.

jvesely: It's probably not the case with show_graph, but rebuilding the complete setup for each test should help avoid transitive/cascading failures.

jvesely: another option would be add a test class and have the constructed composition be a class member accessible in every test function. I think that would also match the show_graph usage pattern.

kmantel: For now I'll leave it as rebuilding - there aren't too many show_graph tests and I don't expect many more will be added - but that's an option if we decide we need to reduce the test suite length in the future.

PR #2089, opened 2021-07-19
github-actions: Drop the old docs compare workflow

jvesely:
Not needed since the new compare comment workflow is merged to master.


jvesely:
there will be other cross-fork PRs to test the fix from #2109, closing.

PR #2109, opened 2021-08-19
ci: fix missing PR number

kmantel:
workflows started by other workflows cannot access the PR
number of the original if the PR branch is from a fork. This causes
compare-comment to not be able to download pnl-ci-docs artifacts

See https://github.com/PrincetonUniversity/PsyNeuLink/runs/3317202966?check_suite_focus=true corresponding to #2103

### DIFF
--- a/.github/workflows/compare-comment.yml
+++ b/.github/workflows/compare-comment.yml
@@ -32,11 +32,14 @@ jobs:
           core.setOutput('DOCS_GEN_ENV', docsSuffix);
 
           var docsArtifacts = artifacts.data.artifacts.filter((artifact) => {
-            return artifact.name.endsWith(docsSuffix) && artifact.name.startsWith('Documentation-')
+            return (
+              (artifact.name.endsWith(docsSuffix) && artifact.name.startsWith('Documentation-'))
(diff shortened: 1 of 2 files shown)


jvesely (approved):

jvesely:
Did you test this locally, or is #2089 the expected test case?

kmantel:
It's been tested for branches of the same repo, so it shouldn't break anything that currently works. It hasn't been tested for what it's meant to fix - PRs across forks - but since the artifacts are otherwise uploaded and downloaded as they should be, I don't see why it won't work as intended in that case. #2089 is a good example though, and compare-comment working properly there would show that PR is safe to merge.

jvesely:
> It's been tested for branches of the same repo, so it shouldn't break anything that currently works. It hasn't been tested for what it's meant to fix - PRs across forks - but since the artifacts are otherwise uploaded and downloaded as they should be, I don't see why it won't work as intended in that case. #2089 is a good example though, and compare-comment working properly there would show that PR is safe to merge.

I think same-repo comments working is good enough.
#2089 might need the change to get all the way to master. iirc the triggered workflows are sourced from the default branch

PR #2132, opened 2021-10-01
UDF improvements

kmantel:
- Support passing parameters into UDF init as normal keyword arguments 
- Support custom functions defined as basic string expressions (no function calls)
- Support integration for one custom parameter using the `stateful_parameter` argument

### DIFF
--- a/psyneulink/core/components/component.py
+++ b/psyneulink/core/components/component.py
@@ -1099,8 +1099,6 @@ class Component(JSONDumpable, metaclass=ComponentsMeta):
         Note: if parameter_validation is off, validation is suppressed (for efficiency) (Component class default = on)
 
         """
-        self._handle_illegal_kwargs(**kwargs)
-
         context = Context(
             source=ContextFlags.CONSTRUCTOR,
             execution_phase=ContextFlags.IDLE,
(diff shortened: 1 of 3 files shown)


jvesely:
This looks like an attempt to have a custom python-like parser. Using the `ast` module is much more robust and extensible.
for example
```
class IdVisitor(ast.NodeVisitor):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.ids = []
        self.calls = False

    def visit_Name(self, node):
        self.ids.append(node.id)

    def visit_Call(self, node):
        self.calls = True


v1 = IdVisitor()
v1.visit(ast.parse('a +b * c - 4'))
print(v1.ids)
print(v1.calls)

v2 = IdVisitor()
v2.visit(ast.parse('a +b * c(d) - 4 '))
print(v2.ids)
print(v2.calls)
```
running the above gives:
```
['a', 'b', 'c']
False
['a', 'b']
True
```

kmantel:
Thanks for pointing that out. I've rewritten this using ast parsing, and also have included support for function calls in expressions, which can be present in MDF models.

PR #2133, opened 2021-10-01
Use IntFlag for cost functions

jvesely:
Add test.


kmantel (approved):
Good, I think I remember an annoyance from combining these despite not being flags

PR #2140, opened 2021-10-06
llvm, mechanism: Use Parameter class to build a list of needed parameter ports

jvesely:
Names of Parameter and ParameterPort don't have to match if there are
multiple parameters with the same name. (Like 'seed' in DDM with DDI
function).
Add test modulating seed of DDI within DDM.

### DIFF
--- a/psyneulink/core/components/mechanisms/mechanism.py
+++ b/psyneulink/core/components/mechanisms/mechanism.py
@@ -2916,16 +2916,17 @@ class Mechanism_Base(Mechanism):
         builder.store(builder.load(params_in), params_out)
 
         # Filter out param ports without corresponding params for this function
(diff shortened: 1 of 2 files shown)


kmantel:
> > > @kmantel, is there a better way to go from `ParamterPort` back to the associated parameter than just remembering the original name?
> > 
> > 
> > Yes - the source attribute:
> > ```
> > >>> import psyneulink as pnl
> > >>> m = pnl.LCAMechanism()
> > >>> m.parameter_ports['offset-combination_function'].source is m.combination_function.parameters.offset
> > True
> > ```
> 
> thanks! works much better than my original solution. Not sure why I missed it.

I realized it's not documented, so it's easy to miss. I'll add that soon.

PR #2154, opened 2021-10-26
MDF compatibility

kmantel:
Support for most of MDF v0.3.1, includes

- basic compositions 
- basic mechanisms using a single main function and a single stateful parameter
- projections between mechanisms with non-identity matrices of varying shapes
- python `math` functions, MDF library functions, string expressions as `UserDefinedFunction`s

does not include
- MDF edge weights
- secondary functions (ex: Mechanism integrator functions) (will be supported in next version of MDF)
- noise (will be supported in next version of MDF)

GHA workflow changes:
- increased fetch depth and fetching of tags to ensure `psyneulink` version is detected for `modeci_mdf` cross-dependency

### New requirements
`modeci_mdf` (excluded in x86 builds)


.github/workflows/pnl-ci.yml
```diff
@@ -41,7 +41,10 @@ jobs:
     - name: Checkout sources
       uses: actions/checkout@v2.3.5
       with:
-        fetch-depth: 10
+        fetch-depth: 200
+
+    - name: Checkout tags
+      run: git fetch --tags
```
jvesely, line 47: why does this need to fetch tags?

.github/workflows/pnl-ci-docs.yml
```diff
@@ -43,16 +43,19 @@ jobs:
...
     - name: Checkout pull base
       uses: actions/checkout@v2.3.5
       if: ${{ matrix.pnl-version == 'base' }}
       with:
-        fetch-depth: 10
+        fetch-depth: 200
         ref: ${{ github.base_ref }}
 
+    - name: Checkout tags
+      run: git fetch --tags
```
jvesely, line 57: I'm not sure I understand the goal here.
The main docs page is built on tag events so it already has the correct tag/version,
the branch docs are currently identified with as e.g. "0+untagged.11.gaabca55".
Is the idea here to to have instead something like "v0.9.1.1-67-gaabca55987" ?
the identifying part (short git hash) is the same.

kmantel: Yes exactly, modeci_mdf depends on a specific psyneulink version, and if the local version of psyneulink is untagged, the installation will fail ([example](https://github.com/kmantel/PsyNeuLink/runs/3902174562?check_suite_focus=true)). I think this is due to the more restrictive dependency checks in pip 20.3. 

If the repo doesn't have older tags and a history from the current commit to one of those tags (reason for fetch-depth change above), versioneer will leave the version as untagged[...]

jvesely: ah, that explains it, thanks! Can you add a short comment around tag fetching step?
so modeci_mdf has circular dependency on pnl?

kmantel: Sure, I will, and yes modeci_mdf depends on psyneulink as well

kmantel:
Supports MDF v0.3.2

PR #2163, opened 2021-10-29
github-actions: Do not check for GITHUB_TOKEN env var when uploading coveralls result

jvesely:


jvesely:
retest this please

PR #2183, opened 2021-11-05
Function_Base: don't crash random_state get with no value in context

kmantel:


jvesely:
This could use explanation on how the value can be None.
The assumption is that is has to be an integer either user specified or retrieved from get_global_seed

kmantel:
Basically, if there is no seed value for the context passed in, getting random_state will crash rather than return None, which is the default behavior for other parameters.

```
>>> t = pnl.TransferMechanism(noise=pnl.UniformDist)
>>> t.integrator_function.noise.base.parameters.random_state.get('ctx')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../PsyNeuLink/psyneulink/core/globals/context.py", line 744, in wrapper
    return func(*args, **kwargs)
  File ".../PsyNeuLink/psyneulink/core/globals/parameters.py", line 1132, in get
    return self._get(context, **kwargs)
  File ".../PsyNeuLink/psyneulink/core/globals/parameters.py", line 1149, in _get
    value = call_with_pruned_args(self.getter, context=context, **kwargs)
  File ".../PsyNeuLink/psyneulink/core/globals/utilities.py", line 1703, in call_with_pruned_args
    return func(*args, **kwargs)
  File ".../PsyNeuLink/psyneulink/core/components/functions/function.py", line 369, in _random_state_getter
    seed_value = [int(seed_param._get(context=context))]
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

if it's preferred to fail with the assert checking for DEFAULT_SEED, we could also check for None there

jvesely:
So is this only for users trying to retrieve values for non-existing context?
Other places depend on the fact that this will return a valid RandomState object and call methods on the returned value.
If "None for non-existent context' is the expected behaviour than the change LGTM.
I guess there are other guarantees that places that call methods on the returned object never use non-existent contexts.

kmantel:
>So is this only for users trying to retrieve values for non-existing context?
That's one, but I ran into it internally. 

>Other places depend on the fact that this will return a valid RandomState object and call methods on the returned value.
Are there checks if the assert fails that it would be easier to just lump this with that? I chose to return None because that's the default for parameters with no getter. I am ok with either but I think both are preferable to the typerror from int cast.

jvesely:
I'm getting a bit lost on whether None is a legal value of seed, or just dummy return for non-existent contexts.
If it's the latter what decides whether to return parameter.default (should be DEFAULT_SEED), or None?

I think it'd be easier if there was a guarantee that parameters never return None (at least internally) to avoid unnecessary branching, but maybe I'm missing some use cases or situations where it is not possible.

kmantel:
> I'm getting a bit lost on whether None is a legal value of seed, or just dummy return for non-existent contexts. If it's the latter what decides whether to return parameter.default (should be DEFAULT_SEED), or None?

It's the latter. The default behavior is at https://github.com/PrincetonUniversity/PsyNeuLink/blob/912f691028e848659055430f37b6c15273c762f1/psyneulink/core/globals/parameters.py#L1155-L1162

and only documented on the Parameter.fallback_default attribute which is probably unclear.

The seed Parameter could set this to True. I'm not sure what the consequences of making fallback_default=True the default behavior would be.

jvesely:
thanks for the explanation. I think setting fallback_default for seed makes sense.
I don't know enough about why the 'return None' beahviour was needed so I can't comment on that.
What were the internal situations that hit the crash that you mentioned?

kmantel:
>I think setting fallback_default for seed makes sense.

sounds good

>What were the internal situations that hit the crash that you mentioned?

specifically 
https://github.com/kmantel/PsyNeuLink/blob/8f06ff9cefa44196fd53c6b5e35fcd1b1929832d/psyneulink/core/components/ports/parameterport.py#L1028-L1030

where I prefer to get a value in context if it exists but use the default if not, which looking at it again, basically is simulating fallback_default=True for everything. Maybe that is the way to go. The behavior was added a few years ago and may not have been extremely important.
I guess it depends on whether getting a default value despite no value existing for a context is more intuitive. One point in favor is your expectation.

PR #2206, opened 2021-11-16
pnl-ci: build docs-only branches

kmantel:
now that jenkins builds are not produced and the required status is on
actions, we need actions to build these

### DIFF
--- a/.github/workflows/pnl-ci.yml
+++ b/.github/workflows/pnl-ci.yml
@@ -10,9 +10,6 @@ on:
     tags-ignore:
       - 'v**'
   pull_request:
-    paths-ignore:
-      - 'docs/**'
-      - 'doc_requirements.txt'
 
 jobs:
   build:


jvesely:
Why do we need to run tests on docs only branches?
The docs are built using the pnl-ci-docs workflow.

kmantel:
> Why do we need to run tests on docs only branches? The docs are built using the pnl-ci-docs workflow.

Because we require the `build` job status to merge in, you can see on #2201 it's blocked because there isn't a successful status for `build`. I don't think github is able to make distinctions between branches like this. I'm not sure of another way to bypass this without running the normal build for these branches other than just allowing force pushing and doing it manually.

jvesely:
> > Why do we need to run tests on docs only branches? The docs are built using the pnl-ci-docs workflow.
> 
> Because we require the `build` job status to merge in, you can see on #2201 it's blocked because there isn't a successful status for `build`. I don't think github is able to make distinctions between branches like this. I'm not sure of another way to bypass this without running the normal build for these branches other than just allowing force pushing and doing it manually.

I see. Sounds like enabling it for pull requests only should suffice, right?

kmantel:
> > > Why do we need to run tests on docs only branches? The docs are built using the pnl-ci-docs workflow.
> > 
> > 
> > Because we require the `build` job status to merge in, you can see on #2201 it's blocked because there isn't a successful status for `build`. I don't think github is able to make distinctions between branches like this. I'm not sure of another way to bypass this without running the normal build for these branches other than just allowing force pushing and doing it manually.
> 
> I see. Sounds like enabling it for pull requests only should suffice, right?

That should be good enough, I can take out the change for push.

jvesely:
looks like this has come up before, without a better solution: https://github.com/actions/runner/issues/952
should we add the main documentation job as a requirement too? (atm. it's `linux-3.7`)

kmantel:
> looks like this has come up before, without a better solution: [actions/runner#952](https://github.com/actions/runner/issues/952) should we add the main documentation job as a requirement too? (atm. it's `linux-3.7`)

I think that's reasonable, I've added the `3.7, x64, ubuntu-latest, pull_request, head` job as required

PR #2242, opened 2021-12-14
dependabot: Do not autorebase pending PRs

jvesely:


jvesely:
This should help reduce CI/CD load and the amount of doc diff comments on PRs.

PR #2279, opened 2022-01-07
OCM: improve simulation context setup

kmantel:
- initialize context for when an agent_rep uses a controller other than
its self.controller
	(ex. test_nested_composition_as_agent_rep)

- add Composition._initialize_as_agent_rep and
Composition._clean_up_as_agent_rep

### DIFF
--- a/psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
+++ b/psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
@@ -1930,9 +1930,18 @@ class OptimizationControlMechanism(ControlMechanism):
 
         # freeze the values of current context, because they can be changed in between simulations,
         # and the simulations must start from the exact spot
-        self.agent_rep._initialize_from_context(self._get_frozen_context(context),
-                                                base_context=context,
-                                                override=True)
+        frozen_context = self._get_frozen_context(context)
+
+        alt_controller = None
+        if self.agent_rep.controller is None:
+            try:
+                alt_controller = context.composition.controller
+            except AttributeError:
+                pass
+
+        self.agent_rep._initialize_as_agent_rep(
+            frozen_context, base_context=context, alt_controller=alt_controller
+        )
 
         # Get control_allocation that optimizes net_outcome using OptimizationControlMechanism's function
         # IMPLEMENTATION NOTE: skip ControlMechanism._execute since it is a stub method that returns input_values
@@ -1945,7 +1954,7 @@ class OptimizationControlMechanism(ControlMechanism):
                                                 )
 
         # clean up frozen values after execution
-        self.agent_rep._delete_contexts(self._get_frozen_context(context))
+        self.agent_rep._clean_up_as_agent_rep(frozen_context, alt_controller=alt_controller)
 
         optimal_control_allocation = np.array(optimal_control_allocation).reshape((len(self.defaults.value), 1))
         if self.function.save_samples:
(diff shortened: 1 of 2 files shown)


psyneulink/core/compositions/composition.py
```diff
@@ -10867,6 +10867,23 @@ def _delete_contexts(self, *contexts, check_simulation_storage=False, visited=No
             except AttributeError:
                 self.scheduler._delete_counts(c)
 
+    def _initialize_as_agent_rep(self, context, base_context, alt_controller=None):
```
jvesely, line 10870: should this assert that only one of `self.controller` and `alt_controller` is `not None`?

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
```diff
@@ -1877,9 +1877,18 @@ def _execute(self, variable=None, context=None, runtime_params=None):
...
-        self.agent_rep._initialize_from_context(self._get_frozen_context(context),
-                                                base_context=context,
-                                                override=True)
+        frozen_context = self._get_frozen_context(context)
+
+        alt_controller = None
+        if self.agent_rep.controller is None:
+            try:
+                alt_controller = context.composition.controller
+            except AttributeError:
```
jvesely, line 1886: is this supposed to catch both missing `context.composition` and `context.composition.controller`?
if it's only the latter.
`alt_controller = getattr(context.composition, 'controller', None)`
would be more readable.

kmantel: It was meant to catch both. I'm unsure if there's a case where this code would be run in a context with no composition set, but if that happened I think it should handle it.

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
```diff
@@ -1954,11 +1972,17 @@ def evaluate_agent_rep(self, control_allocation, context=None, return_results=Fa
...
+                except AttributeError:
+                    pass
             # KDM 5/20/19: crudely using default here because it is a stateless parameter
             # and there is a bug in setting parameter values on init, see TODO note above
             # call to self._instantiate_defaults around component.py:1115
             if self.defaults.search_statefulness:
-                new_context = self._set_up_simulation(context, control_allocation)
+                new_context = self._set_up_simulation(context, control_allocation, alt_controller)
             else:
                 new_context = context
```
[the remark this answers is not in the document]
kmantel: Sorry, I'm not seeing what part you're referencing here. Is it a new change or something in general to simulation contexts?

(I do see the unintentional loss of `check_simulation_storage=True` arg to delete_contexts)

psyneulink/core/compositions/composition.py
kmantel: Actually I'll have to check on what's intended here. If that's true, I'll add the assert

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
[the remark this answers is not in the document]
jvesely: it's not related to the change. the else branch just reuses the base context.
wouldn't the 'value' parameter now include results from simulations stored under the base context?

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
jvesely: is the absence of `composition` in `context` legal in this situation? or is it punting the error to be reported later?

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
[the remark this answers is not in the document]
kmantel: Yes, it's the intended behavior if search_statefulness is set to False. I think @davidt0x had a use case for this

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
kmantel: I think in that case, it's ok and intended not to deal with it. alt_controller would just be one other possibility to try in the case that agent_rep has no controller. If it's not ok for context to have no composition, then I think that should be reported elsewhere.

PR #2280, opened 2022-01-07
Fix/show graph/cim

jdcpni:
• showgraph.py:
  - _assign_incoming_edges():  allow input_CIM of outermost comp to render as node
  - _assign_controller_components():
    allow direct projections from nested nodes to OCM or ObjectiveMechanism when show_cim=False

• test_show_graph.py:
  add test_projections_from_nested_comp_to_ocm_or_obj_mech


tests/composition/test_show_graph.py
```diff
@@ -668,6 +668,106 @@ def test_of_show_nested_show_cim_and_show_node_structure_with_singleton_in_outer
...
+            ):
+                # The test in this condition is for incorrect show_graph output due an as yet undetermined bug
+                #    in which the Projection from the OCM receives a projection to its control_signal output_port
+                #    from a the input_port of a nested monitored node (ic) rather than from its output_port to the
+                #    OCM's outcome_input_port.
+                # If the test fails in this condition, it could mean that the bug has been corrected.
+                # The bug may be the same one as in eb61303808ad2a5ba46fdd18d0e583283397915c
+                raise(AssertionError,"FAILURE TO REPLICATE BUGGY SHOW_GRAPH OUTPUT -- SEE COMMENT IN TEST")
+            else:
+                raise(AssertionError)
```
jvesely: thiscan be just:
`raise`

to forward the existing exception/assertion instead of creating a new one

PR #2294, opened 2022-01-22
llvm, functions/AccumulatorIntegrator: Add compilation support

kmantel:
This is helpful for testing the compiled Threshold condition


jvesely (approved):
Any reason to use a new random scalar for the tests?
Either way, LGTM

kmantel:
> Any reason to use a new random scalar for the tests? Either way, LGTM

I interpreted the random scalars as per parameter rather than as needed per test. I've changed it to use one of the existing ones.

PR #2408, opened 2022-05-06
Revert "requirements: update pytorch requirement from <1.9.0 to <2.0.0"

jvesely:
The versioning scheme uses 1.10.y, 1.11.v, ... instead of bumping to
2.0.x.

This reverts commit 70aa4923ea6f89fe0107206d7f159665a0544675.


jvesely:
We're currently restricted by dependencies to 1.8.0 so the exact version doesn't matter.

PR #2446, opened 2022-07-22
requirements: update torch requirement from <1.9.0 to <1.12.0

jvesely:


jvesely:
The x86 job is failing because pytorch is removed from dependencies on x86
and the package version check can't handle empty dependency.
Ignoring the failure.

PR #2449, opened 2022-07-25
Devel

kmantel:


jdcpni (approved):

jvesely:
Can this wait one more day?
github actions is deprecating macos-10.15 (#2453),
and we should be able to update to numpy 1.21.6 to avoid security warning.

PR #2464, opened 2022-08-12
simple cleanups

jvesely:
Use 'bytes' instead of 'bytearray' for RO view of ctype structures.
Use np.where instead of custom transformation.
User dir(builtins) instead of __builtins__ .


jvesely:
@kmantel are you ok with the 'builtins' change? A longer explanation is in the associated commit message.

kmantel:
Fine by me, it looks like that is the preferred way of using builtin names

kmantel (approved):

PR #2474, opened 2022-08-22
Fix reference leaks and add compiled cleanup destruction check

jvesely:
Clear graph-scheduler caches after every test.
Create a copy of the scheduler condition in composition to avoid leaking a reference to a global instance.
Restore a single run of x86 CI testing.
Add a check for destruction of compiled structures at the end of each compiled test.

### DIFF
--- a/psyneulink/core/compositions/composition.py
+++ b/psyneulink/core/compositions/composition.py
@@ -4097,7 +4097,7 @@ class Composition(Composition_Base, metaclass=ComponentsMeta):
             self.add_controller(controller)
         self.controller_mode = controller_mode
         self.controller_time_scale = controller_time_scale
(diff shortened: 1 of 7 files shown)


psyneulink/core/compositions/composition.py
```diff
@@ -4097,7 +4097,7 @@ def __init__(
             self.add_controller(controller)
         self.controller_mode = controller_mode
         self.controller_time_scale = controller_time_scale
-        self.controller_condition = controller_condition
+        self.controller_condition = copy(controller_condition)
```
kmantel, line 4100: This is better than other ways of dealing with the default

kmantel (approved):
OK by me

PR #2476, opened 2022-08-31
requirements: Update numpy to <1.22.5

jvesely:
Do not check for string values of numpy errors.
Refactor exception handling to check only calls to 'execute'.


kmantel (approved):
lgtm, doesn't seem critical to have the exact numpy error text not change

jvesely:
> lgtm, doesn't seem critical to have the exact numpy error text not change

thank you.
It looks like 1.22 is also hitting some numerical issues which I thought were introduced in 1.23 (this PR passes locally on my Linux setup).
I'll keep this open and rebase after investigating those.

jvesely:
@dependabot rebase

jvesely:
> > lgtm, doesn't seem critical to have the exact numpy error text not change
> 
> thank you. It looks like 1.22 is also hitting some numerical issues which I thought were introduced in 1.23 (this PR passes locally on my Linux setup). I'll keep this open and rebase after investigating those.

the issues are now tracked in #2572

jvesely:
Ignoring results of the x86 CI job. It can't uninstall numpy because it was installed in the restricted package set one step earlier.

PR #2487, opened 2022-09-17
MDF: implement FitzHughNagumoIntegrator

kmantel:


kmantel:
pending next modeci_mdf release that refines function dependency checking

PR #2504, opened 2022-10-20
requirements: update torch requirement from >=1.8.0,<1.12.0 to >=1.8.0,<1.13.0

jvesely:


jvesely:
The failure in x86 is a bug in the CI script that tries to check the version of a package that is not installed.

PR #2521, opened 2022-11-04
Feat/autodiff save

jdcpni:
• autodiffcomposition:
  - add save and load weights from disk


psyneulink/library/compositions/autodiffcomposition.py
```diff
@@ -563,6 +563,27 @@ def execute(self,
                                                         report_num=report_num
                                                         )
 
+    @handle_external_context()
+    def save(self, path:str, context=None):
+        """Saves all parameters to the specified path (e.g. save('my_model.pt'))"""
+        assert (str, "Must provide a path to save the model to!")
```
jvesely, line 569: This is not a correct way to write assertions, both linter and python should complain about it

tests/composition/test_autodiffcomposition.py
```diff
@@ -56,6 +56,59 @@ def test_autodiff_forward(autodiff_mode):
...
+        xor.add_projection(sender=xor_in, projection=hid_map, receiver=xor_hid)
+        xor.add_projection(sender=xor_hid, projection=out_map, receiver=xor_out)
+        return xor
+
+    np.random.seed(0)
+    xor1 = create_xor()
+    xor1_outputs = xor1.run(inputs=[0,0])
+
+    # save
+    xor1.save(tmp_path / 'xor_1.pnl')
```
jvesely, line 93: this is the right way to form file paths. even if it used string concatenation it wouldn't be correct. `os.path.join` is the portable way to combine paths.

tests/composition/test_autodiffcomposition.py
```diff
@@ -56,6 +56,59 @@ def test_autodiff_forward(autodiff_mode):
...
+
+    del xor1
+    pnl.clear_registry()
+
+    # load
+    np.random.seed(1)
+    xor2 = create_xor()
+    xor2_outputs_pre = xor2.run(inputs=[0,0])
+    xor2.load(tmp_path / 'xor_1.pnl')
```
jvesely, line 102: same as above. this is not the way to form file paths.

jdcpni:
Changes incorporated into and superseded by feat/autodiff_various

tests/composition/test_autodiffcomposition.py
jvesely: I was pretty wrong here. the `tmp_path` fixture provides a PathLink instance, so the overloaded `/` operator should do the right thing on all OSes. It explains why it worked the same even after using `os.path.join`. sorry about that

SamKG: No worries!
Good to know, thanks. I'm always surprised by how confusing python portability can be

PR #2531, opened 2022-11-10
docs: Remove invalid escape sequences

jvesely:
The escape sequences were ignored so we can just remove the escaping '\'.
Check that the required warning is present at any position in "test_model_based_num_estimates".
Use raw string when listing code in versioneer.py.
Change warning for invalid escape sequences to error.


jvesely:
the above changes are WIP

PR #2535, opened 2022-11-13
Revert "Add CodeQL workflow for GitHub code scanning"

jvesely:
Reverts PrincetonUniversity/PsyNeuLink#2533
this should go to devel first


jvesely:
Should be fine having this on master. CI complains only if new defects are introduced and the merge established the existing defects on master.

PR #2540, opened 2022-11-18
Nback

jdcpni:
NOTE:  This PR has many changes to learning, as well as the nback script.

• nback.py:
  - executable implementation, though still needs some additional modifications and code support
     to match and validate against version in Beukers et al., 2022:
    - handle biases
    - implement dropout
    - LLVM code for cross entropy as operation in LinearCombination (_gen_llvm_combine)
    - LLVM code for differential of cross entropy in compiledloss.py (_gen_inject_loss_differential)
    - complete integration of statistical analysis of results
    - integrate SweetPea stimulus generation and Kane et al. stimuli
   
• transferfunctions.py:
  - SoftMax: correct derivative computations
  - ReLU: modified to use output (and infer input) if provided, else use input

• learningfunctions.py
  - BackPropagation: 
    - add CROSS_ENTROPY loss
    - fix bug in which derivative for default loss (MSE) was computed using L0

• combinationfunctions.py:
  - LinearCombination: add CROSS_ENTROPY as operation  
    * needs compiled version (see _gen_llvm_combine)

• pytorchmodelcreator.py:
  - _gen_llvm_training_function_body():
    - add support for loss_type = CROSS_ENTROPY
    - modify loss_spec to use keywords for loss

• compiledloss.py:
  - _gen_loss_function: add support for CROSS_ENTROPYLoss
    * needs to be debugged
    * need differential implemented   

• llvm/__init__.py:
  - ExecutionMode:  add ExecutionMode.PyTorch as synonym for ExecutionMode.Python

• composition.py
  - modify loss_spec to use keywords for loss
  - add Report for compiled mode
  - _create_terminal_backprop_learning_components:
    - support loss_function = CROSS_ENTROPY
    - fix bugs in handling of MSE, SSE and L0 loss,  add explicit assignment of objective_mechanism output_ports for each

• autodiffcomposition.py and keywords.py
  - add keywords for loss functions
  - docstring: substantial reorganization
   
• ComparatorMechanism:
  - add L0 and CROSS_ENTROPY standard output_ports

• keywords.py:
  - add Loss(Enum) used for specifying learning
  - replace MSE, SSE, L0, L1, CROSS_ENTROPY, KL_DIV, NLL and POISSON_NLL with Loss members
 
• test_learning.py:
  - test_multilayer_truth:  
    - fix bug in which SSE was assigned as loss, but oputput_port[MSE] was used for objective_mechanism
    - replace with explicit L0 loss and ouput_port[SUM] for objective_mechanism
    -  parameterize to test for L0, SSE and MSE loss


Scripts/Models (Under Development)/Nback/stim/Kane stimuli.py
```diff
@@ -0,0 +1,50 @@
+"""
```
jvesely, line 1: this can use raw string (starts with `r"""` instead of just `"""`) to avoid Python parsing the string.

jdcpni: For the docstring?  (that's the only triple quotes in that file).  If so, why here an not everywhere?

jvesely: It can be used everywhere for docstrings, but changing the existing ones is a bit more involved effort. Python does not parse raw-strings, so the existing Python escapes (`\\`, `\n`, ...) need to be removed.
This is a new docstring so using raw string would get rid of the 'invalid decimal numeral' defect report and probably few 'invalid space sequence' warnings when the file is run.

jvesely: Right, the invalid decimal is probably for the variables, not sure why I see it as coming from the doc string. the part about raw-strings making docs cleaner should still apply.

jdcpni: Got it.  Changed it to r""".  Thanks.

Issue #1703, opened 2020-07-09
Replace typecheck-decorator with python typehinting


kmantel:
typecheck-decorator has not been updated since 2016, and typehints are natively supported in the python `typing` module. python typehints are also supported for automatic documentation with sphinx

jvesely:
typecheck-decorator removed in #2401

PR #2541, opened 2022-11-18
treewide: Cleanup. Reduce the number of warnings issued during test execution

jvesely:
Update patterns that cause Python warnings at runtime
 * Rewrite Angle function algorithm to avoid division and improve performance. Fixes division by zero during initialization.
 * Remove the use of `np.float`. The existing use cases already consider basic `float`, `np.float` is just a deprecated alias of that.
 * Update TransferMechanism tests that override parameters to capture and check for the presence of override warning
 * Update tests that hardcode checks for warnings in a specific order
 * Adjust test configurations to avoid warnings unless the test is intended to exercise situations that produce warnings
 * Use both python 3.7 and python 3.8+ codepaths in AST parser


jvesely:
This removes most "DeprecationWarning:" messages (GA jobs go from ~36k to ~17k). The ones that remain are:
 * coming from typecheck-decorator (should be fixed as part of #1703)
 * coming from onnx
 * use of deprecated `np.matrix`
 * elementwise comparison in `is_compatible`

UserWarnings were reduced to (when running `-m llvm`):
 * missing projection to input port. This is a bug the mechanism function doesn't take input so PNL shouldn't be complaining about a missing projection.
 * default parameters used for GridSearch
 * Compiler shape mismatch warnings.

jvesely:
@kmantel do you mind looking at the change to `is_compatible`?
@jdc do you mind looking at the change to the `Angle` implementation?

psyneulink/core/globals/utilities.py
```diff
@@ -440,7 +440,8 @@ def iscompatible(candidate, reference=None, **kargs):
     try:
         with warnings.catch_warnings():
             warnings.simplefilter(action='ignore', category=FutureWarning)
-            if reference is not None and (candidate == reference):
+            # np.array(...).size > 0 checks for empty list. Everything else create single element (dtype=obejct) array
+            if reference is not None and np.array(candidate, dtype=object).size > 0 and (candidate == reference):
```
kmantel, line 444: No issue with this, but we could potentially also use
```suggestion
            if reference is not None and np.array(candidate, dtype=object).size > 0 and safe_equals(candidate, reference):
```

although safe_equals would need a couple small changes because I think the type of warning produced by that == check has changed since it was written

jvesely: I had the same idea [0,1], but it gave enough errors to leave it for the next time.
Most of the remaining warnings (in total number) are from the elementwise comparison and np.matrix so it's definitely worth revisting.

[0] https://github.com/jvesely/PsyNeuLink/actions/runs/3492618053
[1] https://ci.appveyor.com/project/jvesely/psyneulink-cuda/builds/45417908/job/m8adfgaec8pfs27u

kmantel: Ah ok, makes sense. I can make those changes separately then, they're not many.

PR #2542, opened 2022-11-20
tests: Use custom name prefix for Components instantiated during test enumeration.

jvesely:
The names of components instantiated during test enumeration depend on the order in which the tests are enumerated and shouldn't be used.
Introduce a concept of reserved names (names beginning with "__pnl_" and add a mechanism to add a custom reserved prefix to all generated Component names.
Use the new mechanism to add a common prefix to all Components instantiated during test enumeration.
This makes sure that the names of such components don't conflict with names autogenerated during test execution.
It also makes it easier to detect if any test depends on the name of a Component passed via a parameter.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -36,15 +36,6 @@ def pytest_addoption(parser):
     parser.addoption('--fp-precision', action='store', default='fp64', choices=['fp32', 'fp64'],
                      help='Set default fp precision for the runtime compiler. Default: fp64')
 
-def pytest_sessionstart(session):
-    precision = session.config.getvalue("--fp-precision")
-    if precision == 'fp64':
-        pnlvm.LLVMBuilderContext.default_float_ty = pnlvm.ir.DoubleType()
-    elif precision == 'fp32':
-        pnlvm.LLVMBuilderContext.default_float_ty = pnlvm.ir.FloatType()
-    else:
-        assert False, "Unsupported precision parameter: {}".format(precision)
-
 def pytest_runtest_setup(item):
     # Check that all 'cuda' tests are also marked 'llvm'
     assert 'llvm' in item.keywords or 'cuda' not in item.keywords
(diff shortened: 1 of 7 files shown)


jvesely:
This fixes failure in `test_multiple_projections_to_node_of_nested_composition` when the test is run as the first in the testing session (process).

kmantel:
Is the issue that components are being created in test files - but outside individual tests - so they receive names that are later lost or duplicated somehow because we clear the registry after each test? Are we testing for names of components that were generated during enumeration?

jvesely:
> Is the issue that components are being created in test files - but outside individual tests - so they receive names that are later lost or duplicated somehow because we clear the registry after each test? Are we testing for names of components that were generated during enumeration?

Yes, there are components (~420) that are created during enumeration. I think it's those that are used as parts of 'parametrize'. These components outlive individual tests; they are created before the first test gets executed, and destroyed after the last test finishes.
I think this potentially creates bigger problems; if such a component is assigned to a composition it outlives the test and leaks the entire composition because pytest keeps a reference to the component used in "parameterize".

I don't quite understand the implications of removing components from its registry and then using it in a later test. we've been doing it for some time, so I guess it doesn't break anything during execution.

we're currently not testing for names of components generated during enumeration, but  I think this is more of a coincidence. There's nothing preventing e.g. a show_graph test from using component instances in "parametrize" marks.

I think preventing Components from being isntantiated outside of the test would be a better situation (and also potentially fix some of the leaks). Given that it's rather widespread the pattern is I think it makes sense to try to make things work with Components in "parametrize".

kmantel (approved):

kmantel:
Ok, sounds good then. It does seem like a good idea to keep component instantiations within individual tests, but if we changed those now, I think it's likely some will find their way back eventually, so it's good to have this around in that case.

>I don't quite understand the implications of removing components from its registry and then using it in a later test. we've been doing it for some time, so I guess it doesn't break anything during execution.

The only remaining use of the registry I can think of is for generating default names, so I'm not very surprised if nothing breaks when they're removed from it.

jvesely:
> Ok, sounds good then. It does seem like a good idea to keep component instantiations within individual tests, but if we changed those now, I think it's likely some will find their way back eventually, so it's good to have this around in that case.

Yeah, we'd need to have quite strong enforcement if we want to go the "no Components outside tests", plus review all the situations that use the pattern now, to see how they can be rewritten to instantiate Components locally.

> > I don't quite understand the implications of removing components from its registry and then using it in a later test. we've been doing it for some time, so I guess it doesn't break anything during execution.
> 
> The only remaining use of the registry I can think of is for generating default names, so I'm not very surprised if nothing breaks when they're removed from it.

makes sense. thanks.

Issue #2581, opened 2023-01-14
test failures when using numpy==1.17.5


jvesely:
pnl sets minimum numpy requirements as >=1.17.0, but the test suite fails with 18 errors:
```
FAILED tests/composition/test_composition.py::TestExecutionOrder::test_exact_time - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_no_dependencies[conditions0-termination_conds0] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions4-termination_conds4] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_no_dependencies[conditions2-termination_conds2] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions1-termination_conds1] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions2-termination_conds2] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions0-termination_conds0] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions5-termination_conds5] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions3-termination_conds3] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
FAILED tests/scheduling/test_condition.py::TestAbsolute::test_TimeInterval_linear_everynms[conditions6-termination_conds6] - TypeError: No loop matching the specified signature and casting was found for ufunc gcd
[...]

PR #2583, opened 2023-01-14
requirements: Bump minimum numpy version to 1.19.0

jvesely:
Calling gcd on list comprehension fails with numpy<1.19.0

Closes: https://github.com/PrincetonUniversity/PsyNeuLink/issues/2581


kmantel (approved):
Looks like this belongs at least on graph-scheduler, I'll go ahead and add that. I suppose it's good to add here too because old scheduler versions could still be used that won't have that bound

Issue #2609, opened 2023-03-09
Potential for false-passing tests due to np.allclose use


kmantel:
For `test_parameter_CIM_routing_from_ControlMechanism`:

https://github.com/PrincetonUniversity/PsyNeuLink/blob/06f3006ab7a0a5e72423267c47db767c004eda3f/tests/composition/test_interfaces.py#L544-L559

The value of `res` above is the output of the last trial run by `ocomp` (`[array([4.])]`). It appears that the test's intent is to verify with tolerance the output of all trials of `ocomp`. This test happens to pass because the input for each trial is the same, and `np.allclose` still returns True as long as each individual numeric value is close, even if the shape is different. So, if this test intended to check all trials and we were to change the inputs, it may incorrectly pass, for example:

```python
res = ocomp.run([[0], [0], [2]]) 
assert np.allclose(res, [[4], [4], [4]])
```
would pass, although `ocomp.results` is `[[array([0.])], [array([0.])], [array([4.])]]`.

`np.allclose` is used extensively in our tests, so regardless of this specific example, it seems possible there are some other tests that are affected. In general, if the intent with a test like this is just to verify the final trial, only a single value should be used to compare for clarity. If the result for all trials is meant to be verified, something like `numpy.testing.assert_array_almost_equal` should be used to ensure the proper assert is happening.

jvesely:
@kmantel should this one be closed?

kmantel:
Yeah, not sure why the github keyword missed it.

PR #2622, opened 2023-04-07
tests: change assert numpy.allclose use to numpy.testing.assert_allclose

kmantel:
`numpy.allclose` attempts to broadcast arguments to the same shape before checking equivalence and also has default tolerances `rtol=1e-5` and `atol=1e-8`. This contrasts with `numpy.testing.assert_allclose`, which fails assertion if there is a shape mismatch, and uses default tolerances `rtol=1e-7` and `atol=0`. This PR changes testing uses of `assert numpy.allclose` to `numpy.testing.assert_allclose` and corrects errors where new assert failures result from just testing code.  

Some uses of `numpy.allclose` remain, and these indicate tests that have some kind of inconsistency with psyneulink behavior. These are to be fixed independently. Where the change of tolerance causes tests to fail, `numpy.testing.assert_allclose` is used with explicitly set tolerances equal to `numpy.allclose` defaults. These can be reviewed separately if needed.

Closes https://github.com/PrincetonUniversity/PsyNeuLink/issues/2609


tests/composition/test_autodiffcomposition.py
```diff
@@ -154,7 +154,7 @@ def test_xor_training_correctness(self, eps, calls, opt, autodiff_mode, benchmar
 
         assert len(results) == len(expected)
         for r, t in zip(results, expected):
```
jvesely, line 156: Can the above assert len, and the zip iteration be replaced by a single `np.testing.assert_allclose`?

tests/composition/test_composition.py
```diff
@@ -3108,12 +3108,12 @@ def test_some_inputs_not_specified(self):
...
 
-        assert np.allclose(A.get_output_values(comp), [[2.0, 4.0], [6.0, 8.0]])
-        assert np.allclose(B.get_output_values(comp), [[3., 6., 9.]])
-        assert np.allclose(C.get_output_values(comp), [[0.]])
-        assert np.allclose(D.get_output_values(comp), [[4.]])
+        np.testing.assert_allclose(A.get_output_values(comp), [[2.0, 4.0], [6.0, 8.0]])
+        np.testing.assert_allclose(B.get_output_values(comp), [[3., 6., 9.]])
+        np.testing.assert_allclose(C.get_output_values(comp), [[0.]])
+        np.testing.assert_allclose(D.get_output_values(comp), [[4.]])
         for i,j in zip(comp.results[0], [[2., 4.], [6., 8.], [3., 6., 9.],[0.], [4.]]):
```
jvesely, line 3115: Can  the zip iteration be replaced by a single `np.testing.assert_allclose`?

tests/functions/test_buffer.py
```diff
@@ -59,7 +59,7 @@ def test_buffer_standalone_noise_function_in_array(self, benchmark):
         expected_val = [[24, 4.693117564500052, 46], [17, 7.744647273059847, 29], [10, 11, 12]]
         for v_v, v_e in zip(val, expected_val):
             for v, e in zip(v_v, v_e):
```
jvesely, line 61: Can the zip iterations be replaced by a single np.testing.assert_allclose?

tests/functions/test_buffer.py
```diff
@@ -83,17 +83,17 @@ def __call__(self):
         expected_val = [[24, 12.0, 46], [17, 12.0, 29], [10, 11, 12]]
         for v_v, v_e in zip(val, expected_val):
             for v, e in zip(v_v, v_e):
```
jvesely, line 85: Can the zip iterations be replaced by a single np.testing.assert_allclose?

tests/functions/test_memory.py
```diff
@@ -499,7 +499,7 @@ def test_DictionaryMemory_add_and_delete_from_memory(self):
                            [[ 1,  2,  3],[100,101,102,103]]]
         for m,e in zip(em.memory,expected_memory):
             for i,j in zip(m,e):
-                assert np.allclose(i,j)
```
jvesely, line 502: Can the zip iterations be replaced by a single call to `np.testing.assert_allclose`?

tests/functions/test_memory.py
```diff
@@ -510,7 +510,7 @@ def test_DictionaryMemory_add_and_delete_from_memory(self):
                            [[ 1,  2,  3],[200,201,202,203]]]
         for m,e in zip(em.memory,expected_memory):
             for i,j in zip(m,e):
-                assert np.allclose(i,j)
```
jvesely, line 513: Can the zip iterations be replaced by a single call to `np.testing.assert_allclose`?

jvesely (commented):
A handful of places uses `zip` iterations to compare multidimensional values where it should be possible to use a single call to `np.assert.all_close` insteead.

tests/composition/test_autodiffcomposition.py
```diff
@@ -454,7 +454,7 @@ def test_semantic_net_training_correctness(self, eps, opt, autodiff_mode, benchm
 
         for res, exp in zip(results, expected):
             for r, e in zip(res, exp):
-                assert np.allclose(r, e)
```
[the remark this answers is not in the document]
kmantel: It should be, it looked correct locally for me so I'll double check this after updating with current devel.

This loop+zips can't be replaced with a single command in this case I think because results and expected have ragged arrays in them, and this causes it to fail with `TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''`

tests/composition/test_autodiffcomposition.py
```diff
@@ -154,7 +154,7 @@ def test_xor_training_correctness(self, eps, calls, opt, autodiff_mode, benchmar
 
         assert len(results) == len(expected)
         for r, t in zip(results, expected):
-            np.testing.assert_allclose(r, t)
```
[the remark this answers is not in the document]
kmantel: Yes, fixing this one.

tests/functions/test_integrator.py
```diff
@@ -194,7 +194,7 @@ def test_execute(func, func_mode, variable, noise, params, benchmark):
     res = benchmark(ex, variable)
     expected = func[1](f.initializer, variable, 3, noise, **params)
     for r, e in zip(res, expected):
-        assert np.allclose(r, e)
```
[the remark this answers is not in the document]
kmantel: Fixing

tests/composition/test_autodiffcomposition.py
kmantel: Fixing

tests/composition/test_composition.py
kmantel: This one can't for the same ragged array reason

tests/functions/test_buffer.py
kmantel: Fixing

tests/functions/test_buffer.py
kmantel: Fixing

tests/functions/test_memory.py
kmantel: No, ragged array

tests/functions/test_memory.py
kmantel: No, also ragged array

PR #2628, opened 2023-04-10
tests: Use np.testing.assert_equal instead of np.all

jvesely:
Assert on the return value of equality.


jdcpni (approved):
Overall looks good to me.  Let me know if there are any particular concerns you had.  Otherwise, will leave it to you do to the merge.

jvesely:
> Overall looks good to me. Let me know if there are any particular concerns you had. Otherwise, will leave it to you do to the merge.

No particular concern, I just noticed that some of the `np.all` calls were unchecked in the tests.
Plus looking for a stricter variant to `assert_allclose` for cases where the results are not subject to rounding.

PR #2636, opened 2023-04-18
Decouple ControlMechanism control_allocation and value

kmantel:
- controlmechanism.py:
  - remove control_allocation as alias of value
  - add control_allocation as property = variables of ControlSignals
  - _apply_control_allocation(): no longer assign control_allocation to value

- optimizationcontrolmechanism.py
  - _execute(): shape optimal_control_allocation based on control_allocation rather than ocm.value
  - _instantiate_control_signals(): call super(), and then self._create_randomization_control_signal()

This PR causes a change in numeric results for Nieuwenhuis2005Model. Inspection of the graphic results by @jdcpni shows them to be sufficiently similar, with the main difference in the initial values. These are attributed to non-concerning initialization changes on this branch. 
    
@kmantel verified that the difference between these new results and prior results begins at approximately 0.0006, quickly decreases to a value of approximately 3.6e-5 by trial 200, then gradually decreases, ultimately settling below 1e-8 by the end of script execution.

This PR ideally should be merged with squash due to some complicated intermediate merges that appeared to have overwritten some unrelated changes on devel.


jvesely:
When you do merge with squash. Please remove all `[skip ci]` from the commit message.

PR #2650, opened 2023-04-28
Python 3.10 Support

davidt0x:
This PR makes a few minor fixes to get things working in Python 3.10. The most notable unresolved issue is that Python 3.10 and PyTorch on windows is not currently supported, see https://pytorch.org/get-started/locally/. Workaround for now is to disable pytorch tests for the windows python 3.10 case.


requirements.txt
```diff
@@ -1,8 +1,7 @@
 autograd<1.6
 beartype<0.14.0
 dill<0.3.7
-elfi<0.8.7
-fastkde>=1.0.19, <1.0.25
+fastkde>=1.0.24
```
jvesely, line 4: This is wrong. Please don't remove the upper bound limit.

davidt0x: Whoops, sorry about that, was unintentional. Should be `fastkde>=1.0.24, <1.0.25` correct?

jvesely: yeah. I assume the >=1.0.24 is specific to Python 3.10, right?
it looks like 1.0.24 still supports all the python versions that pnl does so bumping it globally should be ok

davidt0x:
@jvesely or @kmantel, any idea what is causing these test_notebook failures?

kmantel:
> @jvesely or @kmantel, any idea what is causing these test_notebook failures?

Looks like it's related to the ipython version. They released 8.13.0 earlier today, and it seems that python 3.8 is supported only until 8.12

jdcpni:
What do people think about dropping support for Python 3.8?  We’re now (or about to be) on 3.10, and am I correct that 3.11 is either out or imminent?


jdc



On Apr 28, 2023, at 7:56 PM, Katherine Mantel ***@***.***> wrote:



@jvesely<https://github.com/jvesely> or @kmantel<https://github.com/kmantel>, any idea what is causing these test_notebook failures?

Looks like it's related to the ipython version. They released 8.13.0 earlier today, and it seems that python 3.8 is supported only until 8.12

kmantel:
Looks like a lot of projects have run into the same problem, and it's being tracked here https://github.com/ipython/ipython/issues/14053.

jvesely:
should be fixed for CI in https://github.com/PrincetonUniversity/PsyNeuLink/pull/2651

conftest.py
```diff
@@ -13,7 +14,14 @@
 
 try:
     import torch
-    pytorch_available = True
+
+    # If we are on windows and using Python 3.10, despite it importing correctly, PyTorch is currently broken,
+    # see https://pytorch.org/get-started/locally/ showing lack of support.
```
jvesely, line 19: has this been observed? a rather old issue https://github.com/pytorch/pytorch/issues/90530 suggests that the website just needs to be updated.

jvesely:
This also needs to update `setup.py` to indicate that 3.10 is supported.

conftest.py
davidt0x: Yeah, it isn't working for me locally and it seems to have failed [here](https://github.com/PrincetonUniversity/PsyNeuLink/actions/runs/4832641673/jobs/8611674071) in CI as well. Error is weird though. They have wheels but and things import fine but you get these Numpy not available errors. Couldn't find a work around either.

requirements.txt
davidt0x: OK cap is back.

davidt0x:
Added classifier in setup.py for Python 3.10.

conftest.py
jvesely: I think we're hitting https://github.com/pytorch/pytorch/issues/78341, which was unfortunately closed without a fix.
running python 3.10 locally I see:
```
Python 3.10.11 (tags/v3.10.11:7d4cc5a, Apr  5 2023, 00:38:17) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> torch.tensor([1,2,3])
<stdin>:1: UserWarning: Failed to initialize NumPy: module compiled against API version 0x10 but this version of numpy
 is 0xf . Check the section C-API incompatibility at the Troubleshooting ImportError section at https://numpy.org/devd
ocs/user/troubleshooting-importerror.html#c-api-incompatibility for indications on how to solve this problem . (Trigge
red internally at ..\torch\csrc\utils\tensor_numpy.cpp:84.)
tensor([1, 2, 3])
```

It looks like pytorch packaging is broken. it should either build wheels against the oldest supported numpy wheels or indicate the minimum required version.

PR #2667, opened 2023-05-09
autodiff_composition: Allow using torch 1.11 with python3.10 and Windows

jvesely:
Block all torch wheels on windws+python3.10.
Add compatibility code path for torch <1.12.0.
Check for the availability of torch numpy bindings instead of the Python/OS version.


jvesely:
I'm not 100% sure about the extra dimension in the <1.12.0 path, but it's enough to make the tests pass

davidt0x (commented):
LGTM

PR #2674, opened 2023-05-19
llvm: _comp_cached: handle weakref proxy/ref caching

kmantel:
WeakKeyDictionary stores weak references to its keys, but throws a TypeError and does not dereference a weakref proxy/ref even if the proxied object is the same as one of its keys

### DIFF
--- a/psyneulink/core/llvm/builder_context.py
+++ b/psyneulink/core/llvm/builder_context.py
@@ -23,7 +23,7 @@ import weakref
 
 from psyneulink.core.scheduling.time import Time, TimeScale
 from psyneulink.core.globals.sampleiterator import SampleIterator
-from psyneulink.core.globals.utilities import ContentAddressableList
+from psyneulink.core.globals.utilities import ContentAddressableList, unproxy_weakproxy
 from psyneulink.core import llvm as pnlvm
 
 from . import codegen
@@ -75,6 +75,13 @@ class _node_wrapper():
 def _comp_cached(func):
     @functools.wraps(func)
     def wrapper(bctx, obj):
+        if isinstance(obj, weakref.ProxyTypes):
+            # only call for ProxyTypes because this won't fail on most
+            # objects, but specifically not on 'super()' referenced
+            # below, which would return the original object super() was
+            # called with, resulting in caching the wrong thing here
+            obj = unproxy_weakproxy(obj)
+
         try:
             obj_cache = bctx._cache.setdefault(obj, dict())
         except TypeError:  # 'super()' references can't be cached


kmantel:
This came up when looking at tests/composition/test_autodiffcomposition.py::TestMiscTrainingFunctionality::test_training_then_processing, where AutodiffComposition._get_state_struct_type gets called many times, though it seems intended that the cached result be used

psyneulink/core/llvm/builder_context.py
```diff
@@ -75,6 +75,13 @@ def _gen_llvm_function(self, *, ctx, tags:frozenset):
 def _comp_cached(func):
     @functools.wraps(func)
     def wrapper(bctx, obj):
+        if isinstance(obj, weakref.ProxyTypes):
+            # only call for ProxyTypes because this won't fail on most
+            # objects, but specifically not on 'super()' referenced
+            # below, which would return the original object super() was
+            # called with, resulting in caching the wrong thing here
```
jvesely, line 82: This is a bit confusing. What is the proxy object observed here? and how is it related to autodiff composition?

kmantel: The autodiff relation is incidental - it's only how I happened to notice what seemed like unintentional cache misses. 

The wrapper in `_comp_cached` stores a value for an `obj`, and later is called `weakref.proxy` objects that reference `obj`. This does not locate `obj` in the cache due to `TypeError("cannot create weak reference to 'weakcallableproxy' object")` which is caught https://github.com/PrincetonUniversity/PsyNeuLink/blob/0f3ab10d8ec7577f71af7b243bc3fa7ce88c6562/psyneulink/core/llvm/builder_context.py#L77-L81

and bypasses caching.


The example I found creating these proxies is https://github.com/PrincetonUniversity/PsyNeuLink/blob/0f3ab10d8ec7577f71af7b243bc3fa7ce88c6562/psyneulink/core/llvm/builder_context.py#L64-L66

which was added after `_comp_cached`

jvesely: Right. the change to use `weakref.proxy` (c65e0bd0a617ca87e93c261867c5becd334b182e) introduced the bug in pr #2613. I should have checked the total number of generated structures.
I'm still unsure about supporting proxy object caching vs. just reverting c65e0bd.

Does this change fix the high number of generated structures in `test_training_then_processing`? I'd expect that test (and all autodiff compositions) to run into the 'super()'  issue instead.

kmantel: >Does this change fix the high number of generated structures in test_training_then_processing? I'd expect that test (and all autodiff compositions) to run into the 'super()' issue instead.

Could you let me know how to check this? 
I'm only aware of repeated calls to https://github.com/PrincetonUniversity/PsyNeuLink/blob/06f3006ab7a0a5e72423267c47db767c004eda3f/psyneulink/library/compositions/autodiffcomposition.py#L858-L865

I missed at first that `bctx._cache` is a WeakKeyDictionary - in that case, it doesn't seem like a problem to just call `unproxy_weakproxy` each time to let the super objects be stored as well, but I'm not too sure about the intent or benefits of this cache so I'll defer to you.

jvesely: that's a good question. the comment above mentioned "AutodiffComposition._get_state_struct_type gets called many times" so I thought you had some monitoring set up.
Either way, it prodded me to fix the stats collection for code generation which has been fixed/extended in #2687.
you should be able to get some numbers by enabling printouts via `PNL_LLVM_DEBUG=stat`. running tests might need `-n0` or pytest might hide the output.

the `_comp_cached` wrapper is a generalized caching decorator for binary structure types used by compiled functions. any time there's a call to `get_*_struct_type` it can be cached. There are many repeated calls to get the same structure because the structure construction is often recursive.

specifically "node wrapper" is a pseudo object that represents node and all afferent projections. it gets compiled into a single function and reuses the same data types as composition execute and run. Thus generating node wrapper IR code call composition `get_*_struct_type`.
There are a few places that call `get_.*_struct_type(super())`, so I'm not sure what caching will do with those. It can "just work", but it might need a closer look

PR #2688, opened 2023-06-02
tests: Sanitize PEC tests

jvesely:
Use param id instead of fake parameter.
Drop redundant return statements.
Use full precision expected result.
Skip on CUDA.


jvesely:
cc @davidt0x

jvesely:
The first report of documentation differences is caused by ambiguous references, e.g. "matrix" can refer to several different "matrix" parameters. Without selecting a specific one, Sphinx picks an arbitrary one.

PR #2692, opened 2023-06-07
Fix/default control allocation

jdcpni:
• controlmechanism.py:
  - _instantiate_control_signal_type(): fix assignment of allocation_parameter_default
  - docstring: note to use default_allocation rather than default_variable to set allocation default

• port.py:
  - _gen_llvm_function_body(): got rid of OVERRIDE mismatch workaround, and replaced with test and error.
  
• test_control.py:
  - test_control_of_mech_output_port():
      fix: use shape of specification parameter
           to set default_allocation of ControlMechanism
  - test_hanging_control_spec_outer_controller() and
     test_hanging_control_spec_nearest_controller():
      change: specification of default_variable to default_allocation in ControlMechanisms
  - test_control_signal_default_allocation_specification: fix to accomodate mods above

### DIFF
--- a/psyneulink/core/components/ports/port.py
+++ b/psyneulink/core/components/ports/port.py
@@ -2388,13 +2388,17 @@ class Port_Base(Port):
                 name = None
             elif afferent.sender.modulation == OVERRIDE:
                 # Directly store the value in the output array
-                if f_mod_ptr.type != arg_out.type:
-                    assert len(f_mod_ptr.type.pointee) == 1
-                    warnings.warn("Shape mismatch: Overriding modulation should match parameter port output: {} vs. {}".format(
-                                  afferent.defaults.value, self.defaults.value),
-                                  pnlvm.PNLCompilerWarning)
-                    f_mod_ptr = builder.gep(f_mod_ptr, [ctx.int32_ty(0), ctx.int32_ty(0)])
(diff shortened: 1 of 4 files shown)


psyneulink/core/components/ports/port.py
```diff
@@ -2388,13 +2388,17 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, arg_in, arg_out,
...
             elif afferent.sender.modulation == OVERRIDE:
                 # Directly store the value in the output array
-                if f_mod_ptr.type != arg_out.type:
-                    assert len(f_mod_ptr.type.pointee) == 1
-                    warnings.warn("Shape mismatch: Overriding modulation should match parameter port output: {} vs. {}".format(
-                                  afferent.defaults.value, self.defaults.value),
-                                  pnlvm.PNLCompilerWarning)
-                    f_mod_ptr = builder.gep(f_mod_ptr, [ctx.int32_ty(0), ctx.int32_ty(0)])
-                builder.store(builder.load(f_mod_ptr), arg_out)
+                try:
```
jvesely, line 2391: It should not be possible for user to configure things to fail in this place.
so it should be just `assert f_mod_ptr.type == arg_out.type, "message"`.

jvesely (approved):

PR #2702, opened 2023-06-21
Add Optuna support to PEC

davidt0x:
This PR adds support for Optuna samplers to the PEC. The PEC `optimization_function` argument can now be passed an instance of `PECOptimizationFunction` with the `method` attribute set to an instance of `optuna.samplers.BaseSampler`.


davidt0x:
Failing docs build on 3.7 seems to be related to https://github.com/actions/setup-python/issues/682. I will wait a few days to see if it gets fixed.

PR #2732, opened 2023-07-13
Parameters: handle Parameters as dependencies

kmantel:
Correct bug where a Parameter dependency would fail to be respected if a Parameter P specified its dependency on Parameter Q as a Parameter and not the string 'Q', and Q had a dependency of its own

### DIFF
--- a/psyneulink/core/globals/parameters.py
+++ b/psyneulink/core/globals/parameters.py
@@ -2124,8 +2124,15 @@ class ParametersBase(ParametersTemplate):
         for alias_name in aliases_to_create:
             setattr(self, alias_name, ParameterAlias(name=alias_name, source=getattr(self, alias_name).source))
 
-        for param, value in self.values(show_all=True).items():
+        values = self.values(show_all=True)
+        for param, value in values.items():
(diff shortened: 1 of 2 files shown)


psyneulink/core/globals/parameters.py
```diff
@@ -2124,9 +2124,18 @@ def __init__(self, owner, parent=None):
         for alias_name in aliases_to_create:
             setattr(self, alias_name, ParameterAlias(name=alias_name, source=getattr(self, alias_name).source))
 
-        for param, value in self.values(show_all=True).items():
+        values = self.values(show_all=True)
+        for param, value in values.items():
             self._validate(param, value.default_value)
 
+            if value.dependencies is not None:
```
jvesely, line 2131: just a drive-by comment:
Can this be simplified?
```
value.dependencies = {getattr(d, 'name', d) for d in value.dependencies}
assert all(dep in values for dep in value.dependencies)
```

kmantel: Yes, thanks. I always forget about the default option for getattr.

PR #2753, opened 2023-07-28
tests:ParameterEstimationComposition

jvesely:
Reduce running time.
Provide expected results for different optimization methods


davidt0x (commented):
LGTM

PR #2819, opened 2023-10-17
LogEntries: Do not store references to owner's owner

jvesely:
Not used anywhere


kmantel (approved):
Is it preventing garbage collection somewhere?

jvesely:
> Is it preventing garbage collection somewhere?

Not sure. I'm still debugging issues that prevent https://github.com/PrincetonUniversity/PsyNeuLink/pull/2474 from passing.
This change alone doesn't fix it, so it might not be the culprit, or not the only one. I'm just submitting patches that have been around for some time before they go stale.

PR #2824, opened 2023-10-27
llvm, MemoryFunctions: Implement, use or drop unused parameters

jvesely:
Implement compiled support for NOISE and RATE parameters.
Drop DUPLICATE_KEYS and PREVIOUS_VALUE. Handling duplicate keys is complex and not supported in compiled mode.
Update tests.


jvesely:
This makes all compiled parameters used in the compiled Dictionary implementation.

@kmantel or @jdcpni, could you check the Python implementation of the application of RATE and NOISE?

kmantel (approved):
Looks correct to me

PR #2837, opened 2023-11-12
SoftMax: Switch default one_hot_function to 'None'

jvesely:
A new OneHot instance is created in the constructor if needed. So the default one is either unused or replaced.

### DIFF
--- a/psyneulink/core/components/functions/nonstateful/transferfunctions.py
+++ b/psyneulink/core/components/functions/nonstateful/transferfunctions.py
@@ -3052,7 +3052,7 @@ class SoftMax(TransferFunction):
         bounds = (0, 1)
         output = ALL
         per_item = Parameter(True, pnl_internal=True)
-        one_hot_function = Parameter(OneHot, stateful=False, loggable=False)
+        one_hot_function = Parameter(None, stateful=False, loggable=False)
 
         def _validate_output(self, output):
             options = {ALL, MAX_VAL, MAX_INDICATOR, PROB}


jvesely:
This is noticeable in compiled mode. An extra OneHot function includes a PRNG at ~2.4kB of mutable state.

kmantel:
Hm, I wouldn't expect this change to avoid creating another OneHot. Do you have an example where this is happening? It might be a bug.

I could see where there could be an extra when output is something other than None or ALL - I think in this case the instance created in __init__ would be assigned as one_hot_function.default_value, and then copied into one_hot_function.value

jvesely:
> Hm, I wouldn't expect this change to avoid creating another OneHot. Do you have an example where this is happening? It might be a bug.
> 
> I could see where there could be an extra when output is something other than None or ALL - I think in this case the instance created in __init__ would be assigned as one_hot_function.default_value, and then copied into one_hot_function.value

Would `Parameter(OneHot)` not create an instance in Parameter even if output == ALL ?
I see an instance of OneHot returned when calling get(None) in test_transfer.py[SOFT_MAX] tests

kmantel:
> Would `Parameter(OneHot)` not create an instance in Parameter even if output == ALL ? I see an instance of OneHot returned when calling get(None) in test_transfer.py[SOFT_MAX] tests

Oh ok, you mean one instance is created and you want zero? Then this should be fine for that.

jvesely:
> > Would `Parameter(OneHot)` not create an instance in Parameter even if output == ALL ? I see an instance of OneHot returned when calling get(None) in test_transfer.py[SOFT_MAX] tests
> 
> Oh ok, you mean one instance is created and you want zero? Then this should be fine for that.

Yeah, output == ALL doesn't need OneHot at all, so it should be 0 instances in that case.
Would the instance created by Parameter(OneHot) ever be used? output == ALL doesn't need it, and all other output types create a new instance in constructor and set the parameter value there.

kmantel:
> Yeah, output == ALL doesn't need OneHot at all, so it should be 0 instances in that case. Would the instance created by Parameter(OneHot) ever be used? output == ALL doesn't need it, and all other output types create a new instance in constructor and set the parameter value there.

I think there are a few things going on here. value below is `parameters.one_hot_function.value[None]` / `.one_hot_function` and default_value is `parameters.one_hot_function.default_value` / `defaults.one_hot_function`

1. `output in {None, ALL}`
    1. `Parameter(OneHot)`
        - value: an instance of OneHot
        - default_value: the class OneHot
    2. `Parameter(None)`
        - value: None
        - default_value: None
2. `output not in {None, ALL}`
    1. `Parameter(OneHot)`
        - value: an instance of OneHot
        - default_value: an instance of OneHot
    2. `Parameter(None)`
        - value: an instance of OneHot
        - default_value: an instance of OneHot

I think while this PR will get rid of 1, you might also be seeing 2, which shouldn't be caused by the default value set in Parameters. If it is, then I think that's an unexpected bug. The OneHot created in the constructor should be being copied into default_value. I'm not sure if we need to keep those copies, but that would take some looking into.

PR #2838, opened 2023-11-12
Functions/Stability: Convert 'normalize' to FunctionParameter

jvesely:
Convert Stability.normalize to FunctionParameter.

### DIFF
--- a/psyneulink/core/components/functions/nonstateful/objectivefunctions.py
+++ b/psyneulink/core/components/functions/nonstateful/objectivefunctions.py
@@ -34,7 +34,7 @@ from psyneulink.core.globals.keywords import \
     DEFAULT_VARIABLE, DIFFERENCE, DISTANCE_FUNCTION, DISTANCE_METRICS, DOT_PRODUCT, \
     ENERGY, ENTROPY, EUCLIDEAN, HOLLOW_MATRIX, MATRIX, MAX_ABS_DIFF, NORMALIZE, \
     NORMED_L0_SIMILARITY, OBJECTIVE_FUNCTION_TYPE, SIZE, STABILITY_FUNCTION
-from psyneulink.core.globals.parameters import Parameter, check_user_specified
+from psyneulink.core.globals.parameters import FunctionParameter, Parameter, check_user_specified
 from psyneulink.core.globals.preferences.basepreferenceset import ValidPrefSet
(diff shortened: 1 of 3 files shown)


psyneulink/core/components/functions/nonstateful/objectivefunctions.py
```diff
@@ -205,7 +185,11 @@ class Parameters(ObjectiveFunction.Parameters):
         metric = Parameter(ENERGY, stateful=False)
         metric_fct = Parameter(None, stateful=False, loggable=False)
         transfer_fct = Parameter(None, stateful=False, loggable=False)
-        normalize = Parameter(False, stateful=False)
+        normalize = FunctionParameter(
+            False,
+            function_name='metric_fct',
+            function_parameter_name=NORMALIZE,
```
kmantel, line 191: Specifying function_parameter_name isn't needed if it's the same as the parameter name, but it doesn't bother me if this is more readable.

kmantel (commented):
I'm not sure if normalize and metric were meant to be there like an interface, where they must also be on any hypothetical future ObjectiveFunctions or not. Or if we intend on making any new ObjectiveFunctions at all. Any thoughts @jdcpni?

But I don't see anything wrong here, so if it simplifies something on the LLVM side, OK by me.

jvesely:
> I'm not sure if normalize and metric were meant to be there like an interface, where they must also be on any hypothetical future ObjectiveFunctions or not. Or if we intend on making any new ObjectiveFunctions at all. Any thoughts @jdcpni?
> 
> But I don't see anything wrong here, so if it simplifies something on the LLVM side, OK by me.

I was just an associated cleanup. The part that matters for LLVM is the conversion of normalize to FunctionParameter, since the parameter is not used in the implementation of Stability.

psyneulink/core/components/functions/nonstateful/objectivefunctions.py
jvesely: removed

kmantel:
> I was just an associated cleanup. I've dropped those parts. The part that matters for LLVM is the conversion of normalize to FunctionParameter, since the parameter is not used in the implementation of Stability.

OK, sounds good.

kmantel (approved):

PR #2875, opened 2024-01-04
tests: Seed optuna samplers

jvesely:
Replacing `sampler._rng`, as done in `PECOptimizationFunction`, only works for `optuna.RandomSampler`.


jvesely:
cc @davidt0x

Issue #2874, opened 2024-01-04
Intermittent failures in `test_parameter_optimization_ddm[LLVM-optuna_cmaes_sampler]`


jvesely:
The test fails by selecting the wrong result ~3/100 times.

PRNG determinism in PECOptimizationFunction is done (in a hacky way) by replacing optuna's PRNG instance:
psyneulink/core/components/functions/nonstateful/fitfunctions.py:802
```
                # We need to hook into Optuna's random number generator here so that we can allow PsyNeuLink's RNS to
                # determine the seed for Optuna's RNG. Pretty hacky unfortunately.
                opt_func._rng = np.random.RandomState(self.owner.initial_seed)
```

which is both insufficient and incorrect.

It is insufficient because it only works for instances of `optuna.RandomSampler`. However, this test uses CmaEsSampler.
Moreover only 1 dimension of sampling is used and optuna switches to an embedded RandomSampler, which uses a different PRNG. CmaEsSampler is also using PRNG in a different member;

cmaes instantiates its PRNG in `opt_func._cma_rng` and the independent sampler PRNG is in `opt_func._independent_sampler._rng`.

The hacky assignment also causes issues when updating to optuna 3.4.0 and 3.5.0 since the numpy PRNG instance was moved to _rng.rng leading to the following error:
```
self._rng.rng.uniform(trans.bounds[:, 0], trans.bounds[:, 1])
E       AttributeError: 'numpy.random.mtrand.RandomState' object has no attribute 'rng'
```

The immediate solution to the intermittency is to instantiate optuna samplers with a fixed seed.

The hacky assignment of PRNG should also be removed. If using the same seed as parent OCM (`self.owner.initial_seed`) is required, the `PECOptimizationFunction` needs to instance new optuna samplers using the correct seed.

davidt0x:
Hey @jvesely,

Yeah, I meant to revisit, thanks for catching it. I think I have worked out a better way with this method. The issue is that I want to allow users to pass in instantiated optuna samplers but the RandomState is created when the sampler is created. However, I didn't notice that they have a reseed_rng() method, so I can change the seed and call reseed_rng() to update the actual random state. I think this is cleaner approach. Though, now that I am thinking, it the user passed a seed to the sampler when instantiating it we should probably respect that seed over the PNL seed I guess. I have the proposed change here: https://github.com/PrincetonUniversity/PsyNeuLink/tree/fix_optuna_rng. What do you think?

[...]

PR #2876, opened 2024-01-09
Clean up interface to for specifying seeds to optuna samplers with PEC.

davidt0x:
Some small modifications to PEC\Optuna RNG seeding interface based on suggestions from @jvesely and @jdcpni.

- Hack was removed that set optuna RandomState object under the hood.
- Added ability to pass un-instantiated optuna sampler to `PECOptimizationFunction` and added an argument called `optuna_kwargs` for specifying any desired sampler construction args. This allows use the PEC `initial_seed` if it is specified. 
- Added a warning to the user if they have specified a PEC `initial_seed` with an already instantiated optuna sampler. 

Fixes issue #2874

### DIFF
--- a/Scripts/Debug/stability_flexibility/stability_flexibility_pec_optimize.py
+++ b/Scripts/Debug/stability_flexibility/stability_flexibility_pec_optimize.py
@@ -114,7 +114,7 @@ pec = pnl.ParameterEstimationComposition(
         responseGate.output_ports[0],
     ],
     objective_function=reward_rate,
-    optimization_function=pnl.PECOptimizationFunction(method=optuna.samplers.CmaEsSampler(),
+    optimization_function=pnl.PECOptimizationFunction(method=optuna.samplers.QMCSampler(),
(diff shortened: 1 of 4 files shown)


jvesely:
Should this go straight to master?

I think it's better to drop 871f960 ("Update optuna pin to latest.") from this PR.
The constraint is not a pin, just an upper bound and dependabot produces more informative commits with links to release notes and a summary of changes between versions.
Alternatively you can cherry-pick [0d10fea](https://github.com/PrincetonUniversity/PsyNeuLink/pull/2859/commits/0d10fea49edd4cb5e7375d0490bf0c88e03f9592) to include that information.

davidt0x:
@jvesely, whoops that was a mistake about going straight to master. Thanks for catching. I will drop the change of the pin as well.

jvesely:
One more nit. Will the existing two tests that use instanced samplers issue warnings? If so, should the test check that the warnings are emitted in those two cases?

PR #2878, opened 2024-01-11
tests: MDF: correct use of shared state

kmantel:
globals/locals state was incorrectly shared between reference and test model output due to use of eval/exec, resulting in falsely passing tests

### DIFF
--- a/tests/mdf/test_mdf.py
+++ b/tests/mdf/test_mdf.py
@@ -76,6 +76,57 @@ json_results_parametrization = [
 ]
 
 
+def read_defined_model_script(filename):
+    filename = os.path.join(os.path.dirname(__file__), filename)
+
+    with open(filename, 'r') as orig_file:
+        model_input = orig_file.read()
+
+    return model_input
+
+
+def get_loaded_model_state(model_input: str):
+    _globals = copy.copy(globals())
+    _locals = copy.copy(locals())
+
+    exec(model_input, _globals, _locals)
+
+    return _globals, _locals
+
+
+def run_compositions_in_state(
+    composition_input_strs, _globals, _locals, extra_run_args_str=''
+):
+    results = {}
+
+    for comp_name, inputs in composition_input_strs.items():
+        exec(f'{comp_name}.run(inputs={inputs}, {extra_run_args_str})', _globals, _locals)
+        results[comp_name] = eval(f'{comp_name}.results', _globals, _locals)
+
+    return results, _globals, _locals
+
+
+def get_model_results_and_state(
+    model_input: str, composition_input_strs, extra_run_args_str=''
+):
+    _globals, _locals = get_loaded_model_state(model_input)
+    return run_compositions_in_state(
+        composition_input_strs, _globals, _locals, extra_run_args_str
+    )
+
+
+def assert_result_equality(orig_results, new_results):
+    # compositions
+    assert orig_results.keys() == new_results.keys()
+
+    for comp_name in orig_results:
+        np.testing.assert_allclose(
+            orig_results[comp_name],
+            new_results[comp_name],
+            err_msg=f"Results for composition '{comp_name}' are not equal:"
+        )
+
+
 @pytest.mark.parametrize(
     'filename, composition_name, input_dict_str, simple_edge_format',
     json_results_parametrization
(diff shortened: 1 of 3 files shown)


tests/mdf/test_mdf.py
```diff
@@ -86,21 +137,25 @@ def test_json_results_equivalence(
...
 
     # reset random seed
     pnl.core.globals.utilities.set_global_seed(0)
     # Generate python script from JSON summary of composition and execute
-    json_summary = pnl.generate_json(eval(f'{composition_name}'), simple_edge_format=simple_edge_format)
-    exec(pnl.generate_script_from_json(json_summary))
-    exec(f'{composition_name}.run(inputs={input_dict_str})')
-    new_results = eval(f'{composition_name}.results')
-    assert pnl.safe_equals(orig_results, new_results)
+    json_summary = pnl.generate_json(
```
jvesely, line 151: SHould this use `get_mdf_serialized`? There's a warning that `generate_json` is deprecated.

tests/mdf/test_mdf.py
```diff
@@ -86,21 +137,25 @@ def test_json_results_equivalence(
...
+        eval(f'{composition_name}', orig_globals, orig_locals),
+        simple_edge_format=simple_edge_format
+    )
+    new_script = pnl.generate_script_from_json(json_summary)
```
jvesely, line 155: Should this use `generate_script_from_mdf`? There's a warning that `generate_script_from_json` is deprecated.

tests/mdf/test_mdf.py
```diff
@@ -113,24 +168,29 @@ def test_write_json_file(
...
     # Save json_summary of Composition to file and read back in.
     json_filename = filename.replace('.py','.json')
-    exec(f'pnl.write_json_file({composition_name}, json_filename, simple_edge_format=simple_edge_format)')
-    exec(pnl.generate_script_from_json(json_filename))
-    # exec(f'{composition_name}.run(inputs={input_dict_str})')
-    exec(f'pnl.get_compositions()[0].run(inputs={input_dict_str})')
-    final_results = eval(f'{composition_name}.results')
-    assert pnl.safe_equals(orig_results, final_results)
+    exec(
+        f'pnl.write_json_file({composition_name}, "{json_filename}", simple_edge_format={simple_edge_format})',
```
jvesely, line 185: Should this use `write_mdf_file`? There's a warning that `write_json_file` is deprecated.
Is this writing a new file to the source tree? would it be better to use tmpdir/tmppath fixture and store the temp file elsewhere? If the writing silently fails the test could still pass if there's a stale .json from the previous run, no?

tests/mdf/test_mdf.py
```diff
@@ -148,37 +208,58 @@ def test_write_json_file_multiple_comps(
...
+    )
     # reset random seed
     pnl.core.globals.utilities.set_global_seed(0)
 
     # Save json_summary of Composition to file and read back in.
     json_filename = filename.replace('.py', '.json')
 
-    exec(f'pnl.write_json_file([{",".join(input_dict_strs)}], json_filename)')
+    exec(
+        f'pnl.write_json_file([{",".join(input_dict_strs)}], "{json_filename}")',
```
jvesely, line 223: same here, should this use `write_mdf_file`

tests/mdf/test_mdf.py
```diff
@@ -148,37 +208,58 @@ def test_write_json_file_multiple_comps(
...
+        orig_globals,
+        orig_locals
+    )
     exec(pnl.generate_script_from_json(json_filename))
```
jvesely, line 227: SHould this use `generate_file_from_mdf`?

tests/mdf/test_mdf.py
kmantel: Yes, good point, thanks. Using tempfile module

kmantel:
Replaced deprecated json-named functions with mdf-named functions

tests/mdf/test_mdf.py
jvesely: pytest has some useful fixtures for handling tempfiles: https://docs.pytest.org/en/6.2.x/tmpdir.html

kmantel: Ah ok I see what you meant. Replaced with the tmp_path fixture

PR #2887, opened 2024-01-25
Reduce number of warnings emitted during testing

jvesely:
A collection of minor fixes and updates improves the handling of warnings by updating tests to use new interfaces.
The original distribution of warnings when running 4 test jobs was 7299 emitted warnings in 211 locations:
      5 DeprecationWarning
      9 FutureWarning
      1 MatplotlibDeprecationWarning
      4 PendingDeprecationWarning
     12 PNLCompilerWarning
      2 PytestBenchmarkWarning
      1 PytestRemovedIn8Warning
      1 PytestReturnNotNoneWarning
      4 RuntimeWarning
      1 SyntaxWarning
    171 UserWarning

Changes in this PR reduce this to 5696 emitted warnings in 191 locations:
      2 FutureWarning
      1 MatplotlibDeprecationWarning
      4 PendingDeprecationWarning
     12 PNLCompilerWarning
      2 RuntimeWarning
    170 UserWarning

Individual commits target different warning classes.

### DIFF
--- a/psyneulink/core/components/functions/nonstateful/combinationfunctions.py
+++ b/psyneulink/core/components/functions/nonstateful/combinationfunctions.py
@@ -1442,7 +1442,8 @@ class LinearCombination(
         elif operation == CROSS_ENTROPY:
             v1 = variable[0]
             v2 = variable[1]
-            combination = np.where(np.logical_and(v1 == 0, v2 == 0), 0.0, v1 * np.log(v2))
+            both_zero = np.logical_and(v1 == 0, v2 == 0)
+            combination = v1 * np.where(both_zero, 0.0, np.log(v2, where=np.logical_not(both_zero)))
         else:
             raise FunctionError("Unrecognized operator ({0}) for LinearCombination function".
                                 format(operation.self.Operation.SUM))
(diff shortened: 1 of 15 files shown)


jvesely:
Remaining warnings:
* 2 FutureWarning
Both are issued because of the example in the ParameterPort documentation. I didn't change these to use param.modulated because the documentation explicitly talks about the fact that `mod_*` is available on Mechanism even for parameters of the owned function

* 1 MatplotlibDeprecationWarning
This is also blocking matplotlib version bump (that turns this warning into an error), see #2808, We can't switch to the new name python3.7 is limited to matplotlib 3.5.3, which doesn't have any of the new names.

* 4 PendingDeprecationWarning
All 4 are warnings about using numpy.matrix. These are responsible for the majority of the remaining warning instances (~3200/5696). There was an old effort to remove np.matrix usage, but we can also just silence them until numpy removes support for 'matrix' entirely.

* 12 PNLCompilerWarning
All of them are shape mismatch warnings.

* 2 RuntimeWarning
One is in the already updated CombinationFunction. The one case that remains is in cross_entropy_loss test where only the second vector is 0 resulting in np.log(0).
The second warning is trying to normalize 0 matrix in transferfunctions.py:3960

* 170 UserWarning
Various PNL issued warnings. Most of them should be possible to silence by updating tests to fix the issues they warn about or adding a warning check if a PNL warning is expected.

kmantel (approved):

PR #2897, opened 2024-01-31
Devel

kmantel:


jvesely (approved):

jvesely:
base build of docs will fail (so no docs diff) because of issues with sphinx transitive dependencies that are addressed on devel.

PR #2902, opened 2024-02-03
requirements: Bump pycuda to <2025

jvesely:


jvesely:
compiled GPU tests pass on a local machine

PR #2930, opened 2024-03-19
Remove autograd

davidt0x:
The replaces autograd with PyTorch `torch.func.grad` in GradientOptimization. It also removes references to it in a few other places. There is currently only one test that exercises autograd and this seems to pass using PyTorch.

### DIFF
--- a/psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
+++ b/psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
@@ -32,17 +32,11 @@ from numbers import Number
 
 import numpy as np
 
+# Conditionally import torch
 try:
     import torch
 except ImportError:
     torch = None
-
-# Conditionally import torch, this is a bit more strict than the usual conditional import of pytorch. This is because
-# GradientOptimization needs torch.func.grad, which is not available in pytorch versions less than 2.0.0.
-try:
-    from torch.func import grad
-except ImportError:
-    grad = None
 
 from beartype import beartype
 
@@ -1261,16 +1255,17 @@ class GradientOptimization(OptimizationFunction):
             **kwargs
         )
 
-        # Differentiate objective_function using autograd.grad()
+        # Differentiate objective_function using torch.func.grad()
         if objective_function is not None and not self.gradient_function:
 
             if torch is None:
                 raise ValueError("PyTorch is not installed. Please install PyTorch to use GradientOptimization without "
                                  "specifying a gradient_function.")
 
-            if grad is None:
-                raise ValueError("PyTorch version is too old. Please upgrade PyTorch to >= 2.0 to use "
-                                 "GradientOptimization without specifying a gradient_function.")
+            if 'grad' not in dir(torch.func):
+                raise ValueError("torch.func.grad not found. PyTorch version is probably too old. Please upgrade "
+                                 "PyTorch to >= 2.0 to use GradientOptimization without specifying a "
+                                 "gradient_function.")
 
             try:
                 # Need to wrap objective_function in a lambda to pass to grad because it needs to return a torch tensor
--- a/psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
+++ b/psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
@@ -31,6 +31,15 @@ import warnings
 from numbers import Number
 
 import numpy as np
+
+# Conditionally import torch
+try:
+    import torch
+    from torch.func import grad
+except ImportError:
+    torch = None
+    grad = None
+
 from beartype import beartype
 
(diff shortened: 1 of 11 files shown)


davidt0x:
Hey @jvesely, any idea why this test I marked with `pytest.mark.pytorch` is not getting skipped in CI when pytorch isn't available?

jvesely:
> Hey @jvesely, any idea why this test I marked with `pytest.mark.pytorch` is not getting skipped in CI when pytorch isn't available?

Does the test show when you do:
```
pytest --collect-only -m pytorch
```
?
If not there's an issue in applying marks.

davidt0x:
Yeah, on my local machine it seems to me found. The test is called  <Function test_lvoc_features_function>. It is is the only test in test_control that is marked PyTorch now, before it used autograd. Strange, it seems to be skipped properly when I uninstall pytorch on my machine.

jvesely:
> Yeah, on my local machine it seems to me found. The test is called . It is is the only test in test_control that is marked PyTorch now, before it used autograd. Strange, it seems to be skipped properly when I uninstall pytorch on my machine.

What's the pytest version that you use locally?

davidt0x:
Hey, yeah, I agree, those are good suggestions.  I'll do that. I am running locally pytest 7.2.0

jvesely:
> Hey, yeah, I agree, those are good suggestions. I'll do that. I am running locally pytest 7.2.0

can you retest with pytest-8.0.*, and pytest-8.1.* ? It sounds like a pytest bug (and there were changes to marks), so we ca block the buggy version and report it to maintainers.

psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
```diff
@@ -31,6 +31,19 @@
...
+except ImportError:
+    torch = None
+
+# Conditionally import torch, this is a bit more strict than the usual conditional import of pytorch. This is because
+# GradientOptimization needs torch.func.grad, which is not available in pytorch versions less than 2.0.0.
+try:
+    from torch.func import grad
+except ImportError:
+    grad = None
+
```
jvesely, line 46: this can be moved to else of the above try block

psyneulink/core/globals/utilities.py
```diff
@@ -1026,7 +1022,10 @@ def safe_create_np_array(value):
             # See https://numpy.org/neps/nep-0034-infer-dtype-is-object.html
             try:
                 try:
-                    return np.asarray(value)
+                    if torch and torch.is_tensor(value):
+                        return value
+                    else:
+                        return np.asarray(value)
```
jvesely, line 1028: does this still need to be in a try block?

psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
```diff
@@ -1249,13 +1263,34 @@ def reset(self, default_variable=None, objective_function=None, context=None, **
 
         # Differentiate objective_function using autograd.grad()
         if objective_function is not None and not self.gradient_function:
+
+            if torch is None:
+                raise ValueError("PyTorch is not installed. Please install PyTorch to use GradientOptimization without "
+                                 "specifying a gradient_function.")
+
+            if grad is None:
```
jvesely, line 1271: `if not 'grad' in dir(torch.func):`
should work as well

tests/composition/test_control.py
```diff
@@ -1840,20 +1840,39 @@ def test_lvoc_both_predictors_specs(self):
...
+
+        ocm_kwargs = dict(agent_rep=pnl.RegressionCFA,
+                          state_features=[m1.input_ports[0], m1.input_ports[1], m2.input_port, m2],
+                          state_feature_function=pnl.LinearCombination(offset=10.0),
+                          objective_mechanism=pnl.ObjectiveMechanism(
+                              monitor=[m1, m2]),
+                          function=pnl.GradientOptimization(max_iterations=1),
+                          control_signals=[(pnl.SLOPE, m1), (pnl.SLOPE, m2)])
+
+        try:
```
jvesely, line 1864: `'grad' in dir(torch.func)` should work here as well

psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
davidt0x: Yeah, i like the use of `'grad' in dir(torch.func)`. Changed the code to use that.

psyneulink/core/globals/utilities.py
davidt0x: Nope, I don't think so, moved it out.

psyneulink/core/components/functions/nonstateful/optimizationfunctions.py
davidt0x: Changed

PR #2943, opened 2024-04-16
ci: split windows x86 job

kmantel:
Windows x86 ci runs can fail because the container runs out of memory
during testing. This is caused by memory leaks in psyneulink and limited
available memory on the containers. As a workaround, running the full
set of tests over multiple jobs reduces peak memory usage.

Three jobs replace the original Windows x86 build, running pytest with:
1. -m llvm
2. -m "not llvm and composition"
3. -m "not llvm and not composition"

Jobs with the same OS, python version, and architecture overwrite
uploaded dist packages.

### DIFF
--- a/.github/workflows/pnl-ci.yml
+++ b/.github/workflows/pnl-ci.yml
@@ -68,9 +68,21 @@ jobs:
             extra-args: '--forked -m "not llvm"'
 
           # add 32-bit build on windows
+          # split by marks to reduce peak memory
           - python-version: '3.8'
             python-architecture: 'x86'
             os: windows
+            extra-args: '-m llvm'
+
+          - python-version: '3.8'
+            python-architecture: 'x86'
+            os: windows
+            extra-args: '-m "not llvm and composition"'
+
+          - python-version: '3.8'
+            python-architecture: 'x86'
+            os: windows
+            extra-args: '-m "not llvm and not composition"'
 
           # fp32 run on linux python 3.10
           - python-version: '3.10'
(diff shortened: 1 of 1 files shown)


jvesely:
Does splitting by tag also address the issue?
e.g.
```
          - python-version: '3.8'
            python-architecture: 'x86'
            os: windows
            extra-args: '-m llvm'

          - python-version: '3.8'
            python-architecture: 'x86'
            os: windows
            extra-args: '-m not llvm'
```
(splitting by other tags should work too)

kmantel:
It's a good suggestion - running multiple pytest commands in the job immediately fails if one fails, so the entire suite may not be run.

Is there an additional benefit to splitting by `llvm`? `not llvm` still gives the memory error in that case, so I'd need to make three jobs.

`extra-args: '-m llvm'` 
`extra-args: '-m not llvm and composition'` 
`extra-args: '-m not llvm and not composition'` 

but just splitting by `composition` seems to be ok also

.github/workflows/pnl-ci.yml
```diff
@@ -209,7 +220,7 @@ jobs:
 
     - name: Upload dist packages
       uses: actions/upload-artifact@v4
-      if: matrix.version-restrict == ''
+      if: matrix.version-restrict == '' && (! contains(matrix.python-architecture, 'x86') || ! contains(matrix.extra-args, 'not composition'))
```
jvesely, line 223: does:
```
overwrite: matrix.python-architecture == 'x86'
```
work?
the dist packages should be identical and it would be a bit easier to read

can you also add one sentence about the necessary adjustments to upload to the git commit description

kmantel: It looks like `overwrite` won't allow expressions, only true/false, unless I can't figure how to do it correctly. `overwrite: true` can still work and would currently only affect the x86. If we needed to add other jobs later without overwriting, we'd have to change the name anyway

.github/workflows/pnl-ci.yml
```diff
@@ -181,10 +188,14 @@ jobs:
       timeout-minutes: 180
       run: pytest --junit-xml=tests_out.xml --verbosity=0 -n logical ${{ matrix.extra-args }}
 
+    - name: Set extra-args env for artifacts
+      run: echo "EXTRA_ARGS_DESC=$(echo ${{ matrix.extra-args }} | tr ' ="' '_\-\0')" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
kmantel: I will. It looks like only double quote is actually disallowed, so I can remove the others. But I also think space->underscore is nice although unnecessary

PR #2969, opened 2024-05-17
llvm/cuda: Synchronize CPU and GPU buffers

jvesely:
The values in CPU buffers can change via updates to Python Parameters. GPU buffers thus have to be updated/recreated on every launch.
Similarly, Python Parameters can access the most recent value, so modified buffers have to be downloaded after kernel execution.

Fixes PTXExec and PTXRun failures in the new `test_multiple_runs_with_parameter_change*` tests.


jvesely:
This is just a quick fix. Ideally, GPU work launch would use either managed memory or the `In/Out/InOut` helpers to update CPU side buffers.

PR #2973, opened 2024-05-28
llvm: Consolidate construction and sync of compiled structures

jvesely:
Use a helper function for recursive synchronization of the compiled structures.
Use `ctypes` type instead of value in a call to `sizeof` (`sizeof` builtin types throws an exception).
Exclude Time and Optimizer custom classes from writeback.
Do not invalidate PRNG when syncing compiled seed value.
Make the synchronization methods private, drop explicit synchronization in Function and Mechanism tests.
Restrict shape changes on synchronization to known cases of 0d arrays and 'matrix' Parameters.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -205,24 +205,10 @@ def cuda_param(val):
 @pytest.helpers.register
 def get_func_execution(func, func_mode):
     if func_mode == 'LLVM':
-        ex = pnlvm.execution.FuncExecution(func)
-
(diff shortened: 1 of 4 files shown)


jvesely:
@kmantel, do the changes to seed_setter look ok to you?

kmantel (approved):
It makes sense - there shouldn't be resetting when syncing seed like this

PR #2975, opened 2024-06-04
Add support for Python 3.12

jvesely:
Update embedded versioneer to 0.29 to support Python 3.12
Fix codestyle errors.
Advertise support for Python 3.12
Add Python 3.12 CI jobs.


jvesely:
cc @kmantel . I kept the `root_dir` code in `_version.py`. git history doesn't say when/why it was added. I've checked that it's not coming from the old versioneer (installing the same version also removes it). It's used in showgraph to set the default output directory, so I've re-added it manually.

kmantel:
> cc @kmantel . I kept the `root_dir` code in `_version.py`. git history doesn't say when/why it was added. I've checked that it's not coming from the old versioneer (installing the same version also removes it). It's used in showgraph to set the default output directory, so I've re-added it manually.

I'm seeing it from #1231. I don't think it would have to be in `_version.py` and could be moved to `__init__.py` or somewhere else, but yeah not important for this PR

kmantel (approved):
I also double checked the generated version numbers, and they look ok in 3.12 and down to 3.7

PR #2983, opened 2024-06-21
Add conditional parameter fitting feature to the PEC.

davidt0x:
This PR adds a new parameter to the PEC constructor called `depends_on` that specifies conditions for parameters. This causes the PEC to fit a different parameter for each unique condition value in the dataframe. This PR also adds a new argument called `likelihood_include_mask` which is a 1D bool array which specifies which rows of the data to include in the likelihood calculation. This lets users mask off specific rows in the data from the likelihood calculation. Note however, during estimation for a given parameter this trial will still be executed by the composition.

Implementation of conditional parameter fitting required deep changes for how parameter values are passed to the simulation runs. This is because there was no support in the current compiled implementation for allowing for trial-by-trial variation of the parameters. To accomplish this, control mechanisms are now added to the inner composition for each parameter being estimated. These control mechanisms take the trial-by-trial values for each parameter from the composition input and override the values. Currently, the only parameter that is using the old control allocation approach of modulation is the random seed. A deeper refactoring might make sense in the future.


tests/composition/test_parameterestimationcomposition.py
```diff
@@ -212,25 +257,101 @@ def reward_rate(sim_data):
...
+            raise AssertionError("Did not find warning about random seed")
+    elif isinstance(opt_method, type) and issubclass(opt_method, optuna.samplers.BaseSampler):
+        with pytest.warns(UserWarning) as record:
+            pec.run(inputs=inputs_dict)
 
-# func_mode is a hacky wa to get properly marked; Python, LLVM, and CUDA
-@pytest.mark.composition
-def test_parameter_estimation_ddm_mle(func_mode):
-    """Test parameter estimation of a DDM in integrator mode with MLE."""
+        # Search through the warnings to make sure the one we are looking for is there
```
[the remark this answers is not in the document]
davidt0x: Fixed this, thanks.

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -120,6 +158,11 @@
 @pytest.mark.composition
 @pytest.mark.parametrize("inputs_dict, error_msg", run_input_test_args)
 def test_pec_run_input_formats(inputs_dict, error_msg):
+
+    # Need to convert from mapping proxy
+    # to dict to make the test pass.
+    inputs_dict = {k:v for k,v in inputs_dict.items()}
```
jvesely, line 164: Why is this necessary? can't pec.run handle general mapping type?

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -145,17 +188,29 @@ def test_pec_run_input_formats(inputs_dict, error_msg):
...
     ],
-    ids=["differential_evolution", "optuna_random_sampler", "optuna_cmaes_sampler"],
 )
-def test_parameter_optimization_ddm(func_mode, opt_method, expected_result):
+def test_parameter_optimization_ddm(func_mode, opt_method, optuna_kwargs, expected_result):
     """Test parameter optimization of a DDM in integrator mode"""
 
+    # Dicts are being converted to mappingproxy objects, need to convert to dict.
+    if optuna_kwargs:
+        optuna_kwargs = {k: v for k, v in optuna_kwargs.items()}
```
jvesely, line 212: Why is this necessary? Can't optuna handle general mapping type?

tests/composition/pec/test_parameterestimationcomposition.py
davidt0x: Ok, so PEC was modifying the inputs dict in place, I have made it so that it generates a copy now. I was also having issues with Composition._parse_input_dict not handling mappingproxytype well because it checks if type() == dict. Copying the dict gets around this as well.

tests/composition/pec/test_parameterestimationcomposition.py
davidt0x: Ok, I tried to fix this one better as well. It was a bit trickier to track down. It seems that optuna_kwargs argument gets cached in _user_specified_args since its and arg to a function. Then later, a deepcopy operation fails on the PECOptimizationFunction because it can't copy\pickle the mappingproxytype. I added a change in check_user_specified that converts MappingProxyTypes to dicts, does this look like an appropriate fix @kmantel @jvesely

kmantel: Hm, I think I intended to have `_user_specified_args` be deleted after initialization but it doesn't look like it is. It shouldn't be needed after that. Is it just that one attribute that's a problem?

davidt0x: This seems to have worked for my test, checking to see if it breaks something else.

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -228,14 +273,171 @@ def reward_rate(sim_data):
...
+    params_cond1 = dict(
+        threshold=0.7,
+    )
+
+    params_cond2 = dict(
+        threshold=0.3,
+    )
+
+    comp, data_cond1 = _run_ddm_with_params(**{**ddm_params, **params_cond1}, trial_inputs=trial_inputs)
+    _, data_cond2 = _run_ddm_with_params(**{**ddm_params, **params_cond2}, trial_inputs=trial_inputs)
```
jvesely, line 384: The intermediate dict should not be necessary.
```
_run_ddm_with_params(**ddm_param, **params_cond2, ...)
```

psyneulink/core/components/functions/nonstateful/fitfunctions.py
```diff
@@ -306,6 +319,8 @@ def __init__(
         **kwargs,
     ):
         self.method = method
+        self._optuna_kwargs = {} if optuna_kwargs is None else optuna_kwargs.copy()
```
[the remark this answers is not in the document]
davidt0x: Fixed thanks!

psyneulink/core/compositions/parameterestimationcomposition.py
```diff
@@ -460,12 +472,14 @@ def __init__(
...
         model: Optional[Composition] = None,
         data: Optional[pd.DataFrame] = None,
+        likelihood_include_mask: Optional[np.ndarray] = None,
         data_categorical_dims=None,
         objective_function: Optional[Callable] = None,
         num_estimates: int = 1,
         num_trials_per_estimate: Optional[int] = None,
         initial_seed: Optional[int] = None,
         same_seed_for_all_parameter_combinations: Optional[bool] = None,
+        depends_on: Optional[Dict] = None,
```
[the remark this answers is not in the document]
davidt0x: Fixed thanks!

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -228,14 +273,171 @@ def reward_rate(sim_data):
...
+    if isinstance(opt_method, optuna.samplers.RandomSampler):
+        with pytest.warns(UserWarning, match="initial_seed on PEC is not None, but instantiated optuna sampler is being used."):
+            pec.run(inputs=inputs_dict)
+
+    elif isinstance(opt_method, type) and issubclass(opt_method, optuna.samplers.BaseSampler):
+        with pytest.warns(UserWarning, match="Overriding seed passed to optuna sampler with seed passed to PEC."):
+            pec.run(inputs=inputs_dict)
+
+    else:
+        pec.run(inputs={comp: trial_inputs})
```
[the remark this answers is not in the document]
davidt0x: Nice, I like this a lot, much cleaner. Thanks!

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -228,14 +273,171 @@ def reward_rate(sim_data):
...
+
+    if expected_result is not None:
+        if opt_method == "differential_evolution":
+            np.testing.assert_allclose(
+                list(pec.optimized_parameter_values.values()), expected_result, atol=1e-2
+            )
+        else:
+            np.testing.assert_allclose(list(pec.optimized_parameter_values.values()), expected_result)
```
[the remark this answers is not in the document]
davidt0x: Fixed.

tests/composition/pec/test_parameterestimationcomposition.py
davidt0x: Fixed.

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -228,14 +273,171 @@ def reward_rate(sim_data):
...
+
+
+def test_parameter_estimation_ddm_cond(func_mode):
+
+    if func_mode == "Python":
+        pytest.skip(
+            "Test not yet implemented for Python. Parameter estimate is too slow."
+        )
+
+    def _run_ddm_with_params(
```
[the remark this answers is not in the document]
davidt0x: Whoops, forgot to delete that.

psyneulink/core/compositions/parameterestimationcomposition.py
```diff
@@ -815,9 +903,37 @@ def run(self, *args, **kwargs):
...
+        if inputs:
+
+            # Don't check inputs if we are within a call to evaluate_agent_rep, the inputs have already been checked and
+            # cached on the PEC controller.
+            if ContextFlags.PROCESSING not in context.flags:
+                self.controller.check_pec_inputs(inputs)
+
+            # Copy the inputs so we don't modify the original dict, note, we can't copy the keys because they
+            # are object\mechanisms that are in the underlying composition.
+            inputs = {k: v.copy() for k, v in inputs.items()}
```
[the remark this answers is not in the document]
davidt0x: Yeah, the values, at least in the case that they are numpy arrays already, will be modified in place I believe when construct the parameter values for the control signals. See this [method](https://github.com/PrincetonUniversity/PsyNeuLink/blob/efa28752d0db751d28edcd068c91b8f8f18a7733/psyneulink/core/compositions/parameterestimationcomposition.py#L1208). They will probably get copied in this code (depending on the branch they go down), but I thought it was safer to copy them explicitly here just in case one of the branches leads to modifying them in place.

PR #3018, opened 2024-07-28
requirements: Update scipy requirements from <1.12 to >=1.7.3, < 1.15.0

jvesely:
Set the minimum SciPy version to the last version that supports Python 3.7.
Adjust expected results in the differential evolution test based on SciPy version.
Increase upper bound from <1.12 to <1.15.


jvesely:
cc @davidt0x

PR #3051, opened 2024-09-27
docs: add quotation to pip install

younesStrittmatter:
`pip install example[example]` doesn't work in all shells but `pip install "example[example]"` works consistently.


kmantel:
Thanks!

PR #3073, opened 2024-10-18
Drop support for python 3.7

jvesely:
Bump minimum rich requirement.
Restrict 32-bit job to last available wheels.
Drop Python3.7 CI jobs.
Bump minimum Python requirement to 3.8

### DIFF
--- a/.github/actions/install-pnl/action.yml
+++ b/.github/actions/install-pnl/action.yml
@@ -67,6 +67,8 @@ runs:
           echo "contourpy<1.1.0" >> env_constraints.txt
           # pillow >= 10.0.0 doesn't provide win32 wheel
           echo "pillow < 10.0.0" >> env_constraints.txt
+          # pandas >= 2.1.0 doesn't provide win32 wheel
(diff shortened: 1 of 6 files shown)


davidt0x (approved):
LGTM

kmantel (approved):

jvesely:
I planned to delay this change until https://github.com/PrincetonUniversity/PsyNeuLink/pull/3071 is submitted to have one last release with 3.7 support.
Let me know if it'd be better to have this before the release.

kmantel:
I agree one last release seems good. I was delaying #3071 hoping to get something like #3078 in. That copies the 3.7 jobs from other ci currently.

In this case I also think it would be nice to include something for the Condition IDE tooltips ([condition-names](https://github.com/PrincetonUniversity/PsyNeuLink/condition-names)), but that's not fully finalized yet. If 3.7 is blocking something, then I don't think it's important.

kmantel:
I left in the 3.7 job in test-release.yml; that could also be taken out here, unless it makes sense to continue to test the installation on 3.7 there

Issue #2040, opened 2021-05-10
PsyNeuLink fails to install from sdist package


jvesely:
```
$ pip3 install --user psyneulink --no-binary psyneulink
Collecting psyneulink
  Downloading psyneulink-0.8.8.1.tar.gz (1.6 MB)
     |\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588\u2588| 1.6 MB 8.5 MB/s 
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-3g98kp6d/psyneulink/setup.py'"'"'; __file__='"'"'/tmp/pip-install-3g98kp6d/psyneulink/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-3g98kp6d/psyneulink/pip-egg-info
         cwd: /tmp/pip-install-3g98kp6d/psyneulink/
    Complete output (9 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-3g98kp6d/psyneulink/setup.py", line 79, in <module>
        install_requires=get_requirements(),
      File "/tmp/pip-install-3g98kp6d/psyneulink/setup.py", line 18, in get_requirements
        with open(path.join(here, prefix + 'requirements.txt'), encoding='utf-8') as f:
      File "/usr/lib/python3.8/codecs.py", line 905, in open
        file = builtins.open(filename, mode, buffering)
    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-install-3g98kp6d/psyneulink/requirements.txt'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

PR #3078, opened 2024-10-23
ci: update on-tag pypi release action

kmantel:


jvesely:
The main problem with the workflow is that the GA security model doesn't allow jobs to depend on other jobs that include secrets. So the 'test-release' part cannot depend on 'create-python-dist', as long as the latter uses secrets to upload the packages. Either the upload needs to split off to its own job, or we need to find a way to work around the restriction.
There also isn't a good way to install pnl wheel on a 32bit Python. The pnl-test CI jobs hack around the torch requirement by removing it (and mdf) from the requirements file, but this is not possible for wheels/packages. we'd need to add something like a `notorch` extra if we want wheels/sdist working on 32-bit Python

The idea behind this test workflow was to check if the sdist/wheel packages were created correctly (to catch issues like https://github.com/PrincetonUniversity/PsyNeuLink/issues/2040). I don't think we need the full copy of the special jobs from the main test files. Since we release universal wheel, we should be good enough to just check that both packages install OK on all supported platofrms

Not sure if there's a quick fix for the existing release scripts. I used to test them on a local repo creating test tags (and skipping uploads), because of GA security constraints.

It might be worth using this opportunity to adjust how CI should work combined with release scripts. 
The regular CI jobs can install from wheels instead of editable local folder as is done currently. The wheels are already created in every pnl-test CI job so the steps only need to be shuffled around so the wheels package is used during testing.
The release script could then just test that installation via both sdist and wheel works and they both install the same files. It can just wait for the regular CI job on master instead of re-running the tests.

kmantel:
The logs from the previous run have expired, but I do see "Skip output 'sdist' since it may contain secret.". On my fork, it doesn't seem to have a problem with this https://github.com/kmantel/PsyNeuLink/actions/runs/11527544088. Is that what you meant by how you tested it? Where it's possibly different repo or secrets settings that are letting it through. Or maybe the security model has changed since this was last looked at?

I can see the benefit of reworking CI to use wheels; if shifting to that, I think I'd just go ahead with the next release first without any CI changes.

jvesely:
> The logs from the previous run have expired, but I do see "Skip output 'sdist' since it may contain secret.". On my fork, it doesn't seem to have a problem with this https://github.com/kmantel/PsyNeuLink/actions/runs/11527544088. Is that what you meant by how you tested it? Where it's possibly different repo or secrets settings that are letting it through. Or maybe the security model has changed since this was last looked at?
> 
> I can see the benefit of reworking CI to use wheels; if shifting to that, I think I'd just go ahead with the next release first without any CI changes.

Thanks. hm, it looks like the rules now allow the existing flow. I think it makes sense to fix/resurrect the script independently of changes to the main CI jobs. I assumed that the test-release job will fail, but if it runs, it should do something useful.
Without syncing the special jobs in the matrix, LGTM.

.github/workflows/test-release.yml
```diff
@@ -69,9 +69,60 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        dist: [wheel, sdist]
```
jvesely, line 74: 'dist' might still be needed

.github/workflows/test-release.yml
```diff
@@ -69,9 +69,60 @@ jobs:
...
+        python-version: ['3.7', '3.11', '3.12']
+        python-architecture: ['x64']
+        extra-args: ['']
+        os: [ubuntu, macos, windows]
+        version-restrict: ['']
```
jvesely, line 76: 'version-restrict' is not used in this job

.github/workflows/test-release.yml
```diff
@@ -69,9 +69,60 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        dist: [wheel, sdist]
+        python-version: ['3.7', '3.11', '3.12']
+        python-architecture: ['x64']
+        extra-args: ['']
```
jvesely, line 74: 'extra-args' are not used in this job

.github/workflows/test-release.yml
```diff
@@ -69,9 +69,60 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        dist: [wheel, sdist]
+        python-version: ['3.7', '3.11', '3.12']
+        python-architecture: ['x64']
```
jvesely, line 73: 'python-architecture' is not used in the setup-python action, so it can be dropped from the matrix. It might be a good idea to test macos both aarch64 and x64, but if there's only one arch using the default is IMO sufficient.

.github/workflows/test-release.yml
```diff
@@ -69,9 +69,60 @@ jobs:
...
+        extra-args: ['']
+        os: [ubuntu, macos, windows]
+        version-restrict: ['']
+        include:
+          # code-coverage build on macos python 3.9
```
jvesely, line 78: I think we can skip the specialized jobs in release testing.
32-bit and version restrict tests need modifications of the requirements files which is not done in this workflow.

.github/workflows/test-release.yml
kmantel: For the ci in general or for this workflow/wheel-based ci?

.github/workflows/test-release.yml
kmantel: It's in the previous matrix that I'm putting back, and it's used by the `- name: Install wheel` and `- name: Install sdist` steps

PR #3177, opened 2025-01-29
Changes required to support the latest EM Compositon on psnlv

ddelpiano:
- Changes suggested by Katrine to support the EMComposition viewer wise
- grpc from psyneulink are an old version and the viewer uses a more recent version causing a conflict, since these are not used anymore we commented the import (I can do a clean up later if this is fine).

### DIFF
--- a/psyneulink/core/__init__.py
+++ b/psyneulink/core/__init__.py
@@ -10,11 +10,11 @@ from .compositions import *
 from .globals import *
 from .llvm import *
 from .scheduling import *
-from .rpc import *
+# from .rpc import *
 
 __all__ = list(components.__all__)
 __all__.extend(llvm.__all__)
 __all__.extend(compositions.__all__)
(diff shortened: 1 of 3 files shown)


kmantel:
Hi, I think the changes in #3218 should fix the EMComposition in a more general way than what I gave you before, specifically this change: https://github.com/PrincetonUniversity/PsyNeuLink/pull/3218/files#diff-c671c021254f77d644246afac5244674ac47f284228d2942c97bcd5fde82c662

Parameters labeled as `pnl_internal` or not bypassed this problem, but this labeling should be more of a superficial difference that I think shouldn't matter for what you need.

ddelpiano:
Thanks @kmantel , should close this then?

kmantel:
@ddelpiano they don't affect anything related to RPC, which I see is included here. Do we need to upgrade the protobuf (?) package or remove the psyneulink rpc module?

Issue #3205, opened 2025-02-20
Numpy 2.0+ support


jvesely:
This issue tracks adaptations to new behavior in Numpy 2.0
see https://numpy.org/devdocs/numpy_2_0_migration_guide.html.

Moreover, fastkde <2.1 fails with Numpy 2. and fastkde 2.1+ doesn't work with Numpy 1.x (https://github.com/LBL-EESA/fastkde/issues/53), so they need to be updated in sync

Issue #3207, opened 2025-02-20
Numpy 2: List conversion of np.arange uses numpy typed elements


jvesely:
Numpy 1.x:
```
>>> import numpy as np
>>> list(np.arange(1,2,0.1))
[1.0, 1.1, 1.2000000000000002, 1.3000000000000003, 1.4000000000000004, 1.5000000000000004, 1.6000000000000005, 1.7000000000000006, 1.8000000000000007, 1.9000000000000008]
>>> np.arange(1,2,0.1).tolist()
[1.0, 1.1, 1.2000000000000002, 1.3000000000000003, 1.4000000000000004, 1.5000000000000004, 1.6000000000000005, 1.7000000000000006, 1.8000000000000007, 1.9000000000000008]
```

Numpy 2.y:
```
>>> import numpy as np
>>> list(np.arange(1,2,0.1))
[np.float64(1.0), np.float64(1.1), np.float64(1.2000000000000002), np.float64(1.3000000000000003), np.float64(1.4000000000000004), np.float64(1.5000000000000004), np.float64(1.6000000000000005), np.float64(1.7000000000000006), np.float64(1.8000000000000007), np.float64(1.9000000000000008)]
>>> np.arange(1,2,0.1).tolist()
[1.0, 1.1, 1.2000000000000002, 1.3000000000000003, 1.4000000000000004, 1.5000000000000004, 1.6000000000000005, 1.7000000000000006, 1.8000000000000007, 1.9000000000000008
```

This leads to failures in `test_mdf.py` when using `model_with_control.py` because of difference in generated MDF json:

Numpy 1.x:
```
                                   "search_space": [
                                        "SampleIterator([0.1, 0.4, 0.7000000000000001, 1.0000000000000002])",
                                        "SampleIterator([0.1, 0.4, 0.7000000000000001, 1.0000000000000002])"
                                    ],
```

Numpy 2.y:
```
                                    "search_space": [
                                        "SampleIterator([np.float64(0.1), np.float64(0.4), np.float64(0.7000000000000001), np.float64(1.0000000000000002)])",
                                        "SampleIterator([np.float64(0.1), np.float64(0.4), np.float64(0.7000000000000001), np.float64(1.0000000000000002)])"
                                    ],
```

The latter of which fails to convert `SampleIterator` -> `pnl.SampleIterater` and fails the tests:
```
[...]

PR #3219, opened 2025-02-26
deps: Fix issues and enable Numpy 2+

jvesely:
Don't call np.all on a generator.
Use greedy regexp match to remove function/construction call arguments.
Use convert_to_np_array helper in AutodiffComposition.
Explicitly check for empty arrays when converting ContentAddressableMemory weights to integers.

Allow the latest numpy and fastkde packages.

Closes: https://github.com/PrincetonUniversity/PsyNeuLink/issues/3205

### DIFF
--- a/tests/mdf/model_with_control.py
+++ b/tests/mdf/model_with_control.py
@@ -65,7 +65,7 @@ comp.add_controller(
             },
             {
                 pnl.PROJECTIONS: (pnl.THRESHOLD, Decision),
-                pnl.ALLOCATION_SAMPLES: np.arange(0.1, 1.01, 0.3),
+                pnl.ALLOCATION_SAMPLES: (0.1, 0.4, 0.7000000000000001, 1.0000000000000002),
             },
         ],
     )
--- a/psyneulink/core/components/functions/stateful/memoryfunctions.py
+++ b/psyneulink/core/components/functions/stateful/memoryfunctions.py
@@ -393,7 +393,7 @@ def _distance_field_weights_setter(value, owning_component=None, context=None):
     # NOTE: need the following to accommodate various forms of specification (single value, None's, etc)
     #       that are resolved elsewhere
     # FIX: STANDARDIZE FORMAT FOR FIELDWEIGHTS HERE (AS LIST OF INTS) AND GET RID OF THE FOLLOWING
-    test_val = np.array([int(np.array(val).item()) if val else 0 for val in value])
+    test_val = np.array([int(np.array(val).item()) if (arr := np.array(val)).size and arr.item() else 0 for val in value])
     test_val = np.full(len(variable), test_val) if len(test_val) == 1 else test_val
     test_curr_field_weights = np.array([int(np.array(val).item()) if val else 0 for val in current_field_weights])
     test_curr_field_weights = (np.full(len(variable), test_curr_field_weights) if len(variable) == 1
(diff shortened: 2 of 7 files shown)


tests/mdf/model_with_control.py
```diff
@@ -65,7 +65,7 @@
             },
             {
                 pnl.PROJECTIONS: (pnl.THRESHOLD, Decision),
-                pnl.ALLOCATION_SAMPLES: np.arange(0.1, 1.01, 0.3),
+                pnl.ALLOCATION_SAMPLES: (0.1, 0.4, 0.7000000000000001, 1.0000000000000002),
```
kmantel, line 68: If other things are held up by this PR, I might lean in favor of xfailing the failing tests in test_mdf.py instead. This change does make them pass, but the script generated would also crash if run outside of the test environment. This is because of the new `np.float64(`s in #3207 - the numpy import isn't added, but when run by the test, numpy is already imported, which is unintended. I'll need to spend some time reworking these tests, because there might be other preexisting issues unrelated to numpy 2.

jvesely: The fix in this PR is for the presence of nested parentheses. The issue exists irrespective of Numpy 2 and the modified version of this test fails also when using numpy-1.26.
Ideally, this test would use all possible expressions that can be passed to ALLCOATION_SAMPLES; list, arrange, np.array, tuple, ..., to maximize coverage that the test provides.

There might be other issues caused by Numpy 2 that are not caused by existing tests (either by passing accidentally, or not being tested at all), but I'd prefer not to postpone Numpy 2 enablement for those cases.

kmantel: >The issue exists irrespective of Numpy 2 and the modified version of this test fails also when using numpy-1.26.

Yeah, this is basically what I meant as justification for just xfailing it here instead. But I don't have major gripes with proceeding with this as-is

kmantel: Looking further, it does seem like the issue solved here is numpy 2 specific and other problems are tangential. So I'll resolve this to go in as-is

kmantel (approved):

tests/mdf/model_with_control.py
jvesely: thank you. though I'm not sure I understand.
The issue with the short matching of parentheses is not Numpy 2 specific. The modified test using tuple fails on older numpy versions as well. I can split the fix into a separate PR if that's preferable.

Was the suggestion to change the test to include tuple, but not include the fix and mark the test xfail instead, or leave the test as is and mark it xfail on numpy 2 only?

IIUC, there's a second issue with the test forwarding copies of globals() and locals() in `get_loaded_model_state`, as the model should not depend on what is loaded by the test. This makes the calls to np.float work even though they shouldn't. Is that a correct understanding? I can open a new issue for that.

kmantel: I think I was getting mixed up trying to replicate the tuple-mod failure on numpy<2 and couldn't. But I am able to now and see it's independent. I don't think it's necessary to split it because it's also required for numpy 2, but it can be. 

The original suggestion was to leave the test as-is and mark it as xfail because I didn't know how deep was the second issue, which you have the right understanding of. But I think it's pretty limited, and I have a small fix for it, so I think there's no need to xfail. I just hadn't figured a way to test the second issue currently, because I think it's peculiar to the awkward way `SampleIterator` gets saved into mdf as a string, which happens to include `"np."` in numpy 2.

PR #3259, opened 2025-05-06
docs: support (only) sphinx>=5

kmantel:
increase minimum version to satisfy implicit sphinx transitive dependencies from broken_trans_deps.txt

### DIFF
--- a/doc_requirements.txt
+++ b/doc_requirements.txt
@@ -1,3 +1,3 @@
 psyneulink-sphinx-theme<1.2.8.0
-sphinx<4.2.1
+sphinx>=5, <8.2.4
 sphinx_autodoc_typehints<1.16.0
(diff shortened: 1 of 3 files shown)


doc_requirements.txt
```diff
@@ -1,3 +1,3 @@
 psyneulink-sphinx-theme<1.2.8.0
```
jvesely, line 1: Should the min version of `psyneulink-sphinx-theme` be set as well?

kmantel: Sphinx 5 is ok with the previous theme v1.2.5, sphinx 6 needs v1.2.6, and sphinx 7 needs v1.2.7. I wasn't really sure how best to express that here.

jvesely: yeah, sounds like the theme package is a better place to express it in its dependencies.

PR #3260, opened 2025-05-06
Add support for Python 3.13

jvesely:
Adapt test.
Advertise support for Python 3.13.
Add CI workarounds to build onnx-1.17.0.
Enable Python 3.13 CI jobs.


jvesely:
One more issue to fix. This will be rebased when ready.

PR #3350, opened 2025-09-18
ci: fix call to wmic (removed by default in windows-server-2025)

kmantel:
https://learn.microsoft.com/en-us/windows-server/get-started/removed-deprecated-features-windows-server?tabs=ws25


kmantel:
It looks like the jobs here are still using 2022, though my fork seems to have switched to 2025 https://github.com/kmantel/PsyNeuLink/actions/runs/17817698256/job/50660402656

kmantel:
Changes merged with #3354

PR #3358, opened 2025-09-30
Fix and improve RNG seeding in PEC.

### LINKED ISSUE #3224: Bug: PEC uses same seed for noise on every iteration
The PEC appears to be using the same seed for noise on every iteration, rather than using a different seed every iteration. As a result, if the PEC evaluates the log-likelihood for the same parameter set multiple times, it gives identical log-likelihoods whereas simulation noise should make these likelihoods differ slightly.

In the example below I have deliberately set up the PEC to search for 10 iterations over a parameter that does not affect the model's behavior at all. As a result, the PEC should return a likelihood that is at least slightly different on each iteration, due to noise in the DDM. Instead, it returns an identical likelihood 10 times.

```
import psyneulink as pnl
import numpy as np
import pandas as pd
import optuna

# Construct model
input_mech = pnl.TransferMechanism(
    input_shapes=1,
    integrator_mode=True,
    integration_rate=1,
    termination_threshold=1
)

ddm_mech = pnl.DDM(
    function=pnl.DriftDiffusionIntegrator(
        threshold=0.1,
        noise=0.1,
        time_step_size=0.01,
    ),
    reset_stateful_function_when=pnl.AtTrialStart(),
    output_ports=[pnl.DECISION_OUTCOME, pnl.RESPONSE_TIME],
    execute_until_finished=True
)

comp = pnl.Composition()
comp.add_linear_processing_pathway([input_mech, ddm_mech])

input_dict = {
    input_mech: np.ones((100, 1))
}

# Generate data to fit
comp.run(inputs=input_dict, execution_mode=pnl.ExecutionMode.LLVMRun)

data_to_fit = pd.DataFrame(
        np.squeeze(np.array(comp.results))[:, :], columns=["decision", "response_time"]
)

data_to_fit["decision"] = data_to_fit["decision"].astype("category")

# Construct and run PEC
pec = pnl.ParameterEstimationComposition(
    nodes=comp,
    parameters={("termination_threshold", input_mech): np.linspace(1, 10, 10)},
    outcome_variables=[
        ddm_mech.output_ports[0],
        ddm_mech.output_ports[1],
    ],
    data=data_to_fit,
    optimization_function=pnl.PECOptimizationFunction(method=optuna.samplers.TPESampler, max_iterations=10),
    num_estimates=10,
)

pec.controller.parameters.comp_execution_mode.set("LLVM")
pec.controller.function.parameters.save_values.set(True)

ret = pec.run(inputs=input_dict)
```

davidt0x:
This pull request introduces improvements to the randomization and seed generation logic for simulation-based optimization in the parameter estimation and control mechanisms. It refactors how random seeds are handled and propagated, ensures proper warning handling, and updates related tests to match new outputs. The most important changes are grouped below:

### Main Improvements\Fixes

**Randomization and Seed Generation Enhancements:**

* Added a dedicated random number generator (`_seed_rng`) to `OptimizationControlMechanism` and a new method `gen_new_seed_sequence` for generating a sequence of random seeds for simulations. This ensures each simulation receives a new unique seed, improving entropy and control over randomization. [[1]](diffhunk://#diff-309240110e848263f58fb6f5c5a9b01bdfdd4959a5e0a7e585a9dccd637ceaecR1842-R1844) [[2]](diffhunk://#diff-309240110e848263f58fb6f5c5a9b01bdfdd4959a5e0a7e585a9dccd637ceaecL2963-R2972) [[3]](diffhunk://#diff-309240110e848263f58fb6f5c5a9b01bdfdd4959a5e0a7e585a9dccd637ceaecR3787-R3797)
* Updated optimization and simulation functions (`_run_simulations` in `fitfunctions.py` and `_sequential_evaluate` in `optimizationfunctions.py`) to use the new seed sequence for randomization when appropriate, replacing previous logic that incremented start/stop values. [[1]](diffhunk://#diff-eb3b470737056adaa97fd10f8b5ebcc1be91dfd2b6361b9aace28f0a82c5182dR463-R468) [[2]](diffhunk://#diff-b8ea2090af88ed5d70b15ea87be429fd59fb7e4b4f2093553853aed098c29e1cL796-R800)

* Fixed #3224

**Test Updates for New Logic:**

* Updated expected results in several tests to reflect changes in the randomization and seed logic, ensuring that tests pass with the new outputs. [[1]](diffhunk://#diff-b20a8f17389d60e900898be6b421c8a7315a2dab1c302a4078d1458b68c3bc7aL185-R185) [[2]](diffhunk://#diff-b20a8f17389d60e900898be6b421c8a7315a2dab1c302a4078d1458b68c3bc7aL386-R386) [[3]](diffhunk://#diff-b20a8f17389d60e900898be6b421c8a7315a2dab1c302a4078d1458b68c3bc7aL475-R475)

### Miscellaneous Improvements\Fixes

**Warning Handling Improvements:**

* Improved warning checks in simulation likelihood and objective function wrapper logic to more accurately identify and handle custom warning subclasses (`PECObjectiveFuncWarning`, `BadLikelihoodWarning`). Also, warning filters are set to always catch these warnings during differential evolution fitting. [[1]](diffhunk://#diff-eb3b470737056adaa97fd10f8b5ebcc1be91dfd2b6361b9aace28f0a82c5182dL240-R240) [[2]](diffhunk://#diff-eb3b470737056adaa97fd10f8b5ebcc1be91dfd2b6361b9aace28f0a82c5182dL600-R612) [[3]](diffhunk://#diff-eb3b470737056adaa97fd10f8b5ebcc1be91dfd2b6361b9aace28f0a82c5182dR702-R704)

**Utility and Representation Updates:**

* Modified the `SampleSpec` `__repr__` method to display only the custom specification when present, making debugging and logging more concise.


davidt0x:
Hey @jvesely,

Could you take a quick look at this? I am trying to improve the seeding in the PEC stuff and it is failing on fp32. The current approach to changing the seed for each set of simulation is that it was incremented. This was a hold over from the first versions of the PEC, which I never liked because it kind of limits the entropy I think. I have modified things to generate a new set of random integer seeds each simulation. However, it is failing because of different results on fp32 only. I think I traced the issue to how we are representing the seeds as floats instead of integers in `LLVMBuilderContext.get_random_state_ptr`. With fp32 (24‑bit mantissa) large integers (> 2^24 ≈ 16.7M) cannot be represented exactly, so I end up getting slightly different seeds for fp64 vs fp32. I could just limit the seed range to 2^24 for now I guess, but it seems like a hack. I could modify the tests for different values in fp32 and fp64, but I would like to get the same randomization for fp32 vs fp64. Any ideas on a easy fix?

jvesely:
> Hey @jvesely,
> 
> Could you take a quick look at this? I am trying to improve the seeding in the PEC stuff and it is failing on fp32. The current approach to changing the seed for each set of simulation is that it was incremented. This was a hold over from the first versions of the PEC, which I never liked because it kind of limits the entropy I think. I have modified things to generate a new set of random integer seeds each simulation. However, it is failing because of different results on fp32 only. I think I traced the issue to how we are representing the seeds as floats instead of integers in `LLVMBuilderContext.get_random_state_ptr`. With fp32 (24‑bit mantissa) large integers (> 2^24 ≈ 16.7M) cannot be represented exactly, so I end up getting slightly different seeds for fp64 vs fp32. I could just limit the seed range to 2^24 for now I guess, but it seems like a hack. I could modify the tests for different values in fp32 and fp64, but I would like to get the same randomization for fp32 vs fp64. Any ideas on a easy fix?

Seeds are represented as floats because they need to support modulation, and computation using integer types is not supported. Integer types are unlikely to be supported in compilation, as it'd need an assignment of type for different parameters (most of the parameters are currently initialized to integer `0` even though they are used as floats during computation).

A quick fix is to not do:
```
 return self._seed_rng.integers(0, 2**31 - 1, size=num_estimates).tolist()
```

There is no correlation between random sequences and the distance of their respective seeds.
Using random integers instead of `range()` only increases the code complexity for no benefit.

davidt0x:
> There is no correlation between random sequences and the distance of their respective seeds. Using random integers instead of `range()` only increases the code complexity for no benefit.

Yeah, I realize that streams from consecutive seeds are not correlated. However, in the old code, lets say we are doing 200 simulations and our initial seed is 0. The first set of simulations gets seeds 0 - 199, then the next set of simulations get seeds 1 - 200, etc. That means, for these two sets of simulations we only have 1 new set of random values, because 199 of the simulations will share a seed with the last batch. Remember, I am computing aggregate statistics of these batches of sequences. That means if the parameters for the simulations remain the same, I would get 199 identical reaction times and decisions for DDM for example between these two runs. We could of course increment by num_simulations each time to get a completely new set seeds. However, this approach would still lead to different results between fp32 and fp64 if the initial_seed is greater than 2^21. It would also happen if the total number of simulations plus the initial seed value exceeded 2^21. Ideally, I would just generate 200 seeded RNGs in the beginning and then never seed again. The RNG state would be preserved across each run of 200 simulations and I would get different uncorrelated values. Does what I am saying make sense?

jvesely:
I'm not sure I understand all angles. 

1. Seeds above 24bits are rounded when using fp32. This would be extremely difficult to fix properly, as that would imply detection and support for integer types in computation for some of the values used by pnl components. It can be restricted to seeds, but that's just moving the hack to the compilation side (which already has plenty).
EDIT: Ideally, all values would use numpy types so the compiler can use dtypes to derive compiled types, but that would need a stronger separation between user-provided values/model and the values used in computation.

I'm not sure how significant the issue is outside of tests. Even rounded seeds should be unique and produce independent streams. Those would be different from Python and fp64 versions, but should work OK for collecting statistics.
EDIT: Note that this is for keeping the randint to generate seeds. The risk of seed collision (same seed more than once in the sequence) is slightly higher with fp32 rounding. I wouldn't expect this to be too bad, though I haven't run the numbers.

If the above is correct. We can just have a different set of expected results for fp32.
Alternatively, the seed generation can cast the seed values to np.float32 and back to int to get the same sequences for fp32/fp64/python.

2. Workload split between simulations and trials. If each simulation needs a unique seed, the above would restrict the workload to ~16M simulations. Is that limit too low?  if one round on the order of ~1000 seeds, we'd still need 16000 rounds before the seeds get rounded. How long is one round expected to take? 

3. Using random to generate seeds. I still don't see the benefits of using it. Compared to, e.g., incrementing a static counter to get each new seed. This would have the added benefits of "just working" for tests that only exercise the lower end of the counter.
something like this:
```
>>> i = 0
>>> [i := i+1 for x in range(10)]
[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
>>> [i := i+1 for x in range(10)]
[11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
```

or a quite hacky version for fp32 (the cast to `float` is just for nice output). this should still have 99%+ of the 31bit range
```
>>> f = np.float32(16777216.0)
>>> [float(f := np.nextafter(f, np.inf)) for _ in range(10)]
[16777218.0, 16777220.0, 16777222.0, 16777224.0, 16777226.0, 16777228.0, 16777230.0, 16777232.0, 16777234.0, 16777236.0]
>>> [float(f := np.nextafter(f, np.inf)) for _ in range(10)]
[16777238.0, 16777240.0, 16777242.0, 16777244.0, 16777246.0, 16777248.0, 16777250.0, 16777252.0, 16777254.0, 16777256.0]
```

davidt0x:
Yeah, that is a good point about sequential seeds at least keep the randomness the same for the tests, since the range of seeds is low. I think I will go with that and I will just modulo by 2^24 in the 32-bit case. I will just do that so that it doesn't bias seeds in the fp32 case. I don't think there is a way to get anywhere near 32-bits of effective entropy when we cast to fp32. We might be able to squeeze a few more bits of entropy if we go above 2^24 but it would bias the seeds. That being said, this isn't cryptography so I don't think any of this matters :) It is a cool little bug though! Sorry for wasting your time.

psyneulink/core/components/mechanisms/modulatory/control/optimizationcontrolmechanism.py
```diff
@@ -1838,6 +1838,12 @@ def __init__(self,
 
         function = function or GridSearch
 
+        if initial_seed is None:
+            initial_seed = np.random.randint(0, 2**31 - 1)
```
jvesely, line 1842: I'd be better if this used `_get_global_seed()` or similar. There has been a significant effort to avoid perturbing the global state from within PNL. Libraries that do perturb the global state make things difficult for their users.

PR #3361, opened 2025-10-07
llvm:  Add basic support for aggregated OutputPort variables

jvesely:
Add shared variable spec canonicalization function for both Python and compiled execution.
Create on-stack structure to store aggregated arguments.
Add and update tests.

### DIFF
--- a/psyneulink/core/components/mechanisms/mechanism.py
+++ b/psyneulink/core/components/mechanisms/mechanism.py
@@ -1095,7 +1095,7 @@ from psyneulink.core.components.functions.function import FunctionOutputType
 from psyneulink.core.components.functions.nonstateful.transferfunctions import Linear
 from psyneulink.core.components.ports.inputport import DEFER_VARIABLE_SPEC_TO_MECH_MSG, InputPort
 from psyneulink.core.components.ports.modulatorysignals.modulatorysignal import _is_modulatory_spec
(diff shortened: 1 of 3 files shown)


psyneulink/core/components/mechanisms/mechanism.py
```diff
@@ -3027,123 +3027,167 @@ def _gen_llvm_ports(self, ctx, builder, ports, group,
...
+        for spec in canonical_port_spec:
+            param_name, indices = spec
+
+            if param_name == VALUE:
+                base = value
+            elif param_name in self.llvm_state_ids:
+                base = ctx.get_param_or_state_ptr(builder, self, param_name, state_struct_ptr=mech_state)
             else:
-                data = None
+                assert False, "Unsupported variable spec Parameter: {}".format(param.name)
```
kmantel:
```suggestion
                assert False, "Unsupported variable spec Parameter: {}".format(param_name)
```

probably? F821 popped here

kmantel (approved):
Looks good. I like `_canonicalize_port_variable_specification` - it's in line with how I'd like parsing to work elsewhere also

psyneulink/core/components/mechanisms/mechanism.py
jvesely: yeah, thanks for catching it.

PR #3363, opened 2025-10-07
Move to pyproject.toml, uv, and ruff.

davidt0x:
This pull request introduces several key updates to the development workflow, CI configuration, and documentation for the PsyNeuLink repository. The main themes are: modernizing Python tooling and CI, updating Python version support, improving linting practices, and providing comprehensive contributor guidance. Notably, Python 3.8 support is dropped in favor of 3.9+, CI now uses `uv` and `hatch` for faster builds, Ruff replaces Flake8 for linting, and a new `AGENTS.md` guide for coding agents.

**CI and Tooling Modernization**
- Switched CI and packaging from `pip`/`wheel` to `uv` and `hatch` for faster dependency installation and builds, including cache management and updated build steps in `.github/actions/install-pnl/action.yml`. [[1]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8R33-R35) [[2]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L47-R58) [[3]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L101-R125) [[4]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L127-R143)
- Removed manual pip wheel caching in CI workflows, as `uv` now handles caching internally. [[1]](diffhunk://#diff-7fd7c4eea3dd0f5838f546a7ec422814933f2bf332900772445853e59d8bfde1L70-R70) [[2]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL121-R148)

**Python Version Support Updates**
- Dropped Python 3.8 from all CI matrices and test-release workflows; Python 3.9 is now the minimum supported version in CI. [[1]](diffhunk://#diff-7fd7c4eea3dd0f5838f546a7ec422814933f2bf332900772445853e59d8bfde1L26-R26) [[2]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL40-R40) [[3]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL77-R87) [[4]](diffhunk://#diff-191bb5b4e97db48c9d0bdb945dd00e17b53249422f60a642e9e8d73250b5913aL14-R14) [[5]](diffhunk://#diff-191bb5b4e97db48c9d0bdb945dd00e17b53249422f60a642e9e8d73250b5913aL75-R75). This wasn't really required but I figured its time to do so since its been end-of-life for awhile now.

**Linting and Test Improvements**
- Replaced Flake8 with Ruff for linting in both CI and local development; updated `dev_requirements.txt` and `.github/workflows/pnl-ci.yml` accordingly. [[1]](diffhunk://#diff-cdde4d4751f711bb59b74a062edfaa0d244ae969f7a0fbea7db0951f4c1f501fL9-R11) [[2]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL121-R148)
- Updated linting steps to match previous Flake8 behavior using Ruff's rule selection and exclusion features.

**Contributor Documentation**
- Added a detailed `AGENTS.md` file with practical guidelines for code, test, and doc contributions, including environment setup, testing, coding conventions, and validation checklists.

**Miscellaneous and Maintenance**
- Updated dependency constraints for Windows (prefer wheel-bearing `pywinpty` version) and made minor code/manifest cleanups. [[1]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L59-R69) [[2]](diffhunk://#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003L9-L10)
- Improved version handling and import structure to use a canonical `__version__` attribute. [[1]](diffhunk://#diff-7769159e1600b1987e48d35f257b07c455e6ee46d98543b446aed9760e78397cL38-R40) [[2]](diffhunk://#diff-7769159e1600b1987e48d35f257b07c455e6ee46d98543b446aed9760e78397cL53-L55) [[3]](diffhunk://#diff-008dcb3426febd767787b1521f1fe33086313b927ea37eaab86df5fa88a51698L80-R80)
- Minor typing and code style fixes (e.g., `noqa` for F722).

**References:**  
[[1]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8R33-R35) [[2]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L47-R58) [[3]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L59-R69) [[4]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L101-R125) [[5]](diffhunk://#diff-175ed8713a7d118aad2b0dc597bcd399145b159a0cffa219fa60580a628007e8L127-R143) [[6]](diffhunk://#diff-7fd7c4eea3dd0f5838f546a7ec422814933f2bf332900772445853e59d8bfde1L26-R26) [[7]](diffhunk://#diff-7fd7c4eea3dd0f5838f546a7ec422814933f2bf332900772445853e59d8bfde1L70-R70) [[8]](diffhunk://#diff-7fd7c4eea3dd0f5838f546a7ec422814933f2bf332900772445853e59d8bfde1L101-R88) [[9]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL40-R40) [[10]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL77-R87) [[11]](diffhunk://#diff-323305122aff6acb29db4b8662a36c199be89a155c14b3869e23512fd753ff1fL121-R148) [[12]](diffhunk://#diff-191bb5b4e97db48c9d0bdb945dd00e17b53249422f60a642e9e8d73250b5913aL14-R14) [[13]](diffhunk://#diff-191bb5b4e97db48c9d0bdb945dd00e17b53249422f60a642e9e8d73250b5913aL75-R75) [[14]](diffhunk://#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R1-R117) [[15]](diffhunk://#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003L9-L10) [[16]](diffhunk://#diff-cdde4d4751f711bb59b74a062edfaa0d244ae969f7a0fbea7db0951f4c1f501fL9-R11) [[17]](diffhunk://#diff-008dcb3426febd767787b1521f1fe33086313b927ea37eaab86df5fa88a51698L80-R80) [[18]](diffhunk://#diff-7769159e1600b1987e48d35f257b07c455e6ee46d98543b446aed9760e78397cL38-R40) [[19]](diffhunk://#diff-7769159e1600b1987e48d35f257b07c455e6ee46d98543b446aed9760e78397cL53-L55) [[20]](diffhunk://#diff-56cfb361fd4f3472ac9237605646c50845ebf6aabf8fde8ff5b82223dd9ad619L1346-R1346)


jvesely:
This should be split into 3-4 separate PRs.

Few more issues:
* `swap pycodestyle\pydocstyle for ruff ` Why? What does it bring that existing linting doesn't do?
* Why was Python 3.8 dropped? Have there been any issues, or does anything depend on Python 3.9 features?
* `Migrate to uv pip install for install-pnl action.` Why? This appears to be only causing issues:
  * The issue with test enumeration and onnxruntime doesn't appear on the current CI. Why does it appear with this change?
  * The issue with broken Python 3.9 on macOS also doesn't manifest with the current CI. Why does it appear with this change?
  
```
uses: astral-sh/setup-uv@v6
```
third-party actions should be no-go unless their code and release process are reviewed to make sure we won't be stuck with abandoned dependency

davidt0x:
Thanks for taking a look so quick! Yeah, it probably should be split into a couple PRs. I just got carried away trying things. Started with the move to pyproject.toml, then got annoyed that pycodestlye didn't have options to configure from pyproject.toml (wanted to get rid of setup.cfg entirely) and doesn't seem [interested](https://github.com/PyCQA/pycodestyle/issues/813) in adding that feature. Then one thing lead to another :)

To address your points:

* **Why ruff over pycodestyle\pydocstyle?** Overall, it is just a better tool in my opinion. Want to use it locally for this project so I was hoping we could mirror that in CI. Currently, it seems better maintained, superior speed (not too important in CI but when running locally its nice), supports extensive configuration from pyproject.toml. There will probably be some growing pains at first but I think they are worth it long term. We could incorporate pycodestyle, pydocstyle, flake8, and isort checks (and auto-fixes in some cases) all from one tool.  My next plan was to try to see what it could auto-fix for us. One downside is it doesn't seem to support all the extra options for the unused args that flake8 supports (nested functions, etc.), I had to manually noqa some of these. We could still use flake8 for these tests of course since we were just doing it as a separate CI step. I was just trying to simplify things. 

* **Why drop Python 3.8**. Ultimately, I think its time to drop it because it was end-of-life 2024-10. For example, I believe numpy and pytorch have both dropped it. However, I don't really feel strongly about it. I don't think its breaking anything at the moment though so I could put it back. Its more of a why keep it? Eventually it will break something and we will drop it then I figure. Its not like we have a ton of users that we are maintaining right now. If they do come, we should probably do our small part to encourage them to upgrade their python version. I guess this should be a separate PR at least.

* **uv migration**. uv is nice tool, once again I don't feel too strongly about this one, I could go either way. I have been using it locally and I see a lot of other folks moving to it. It seems to be the general direction for the python ecosystem. I think it would be good to test using it at least. It saves us 30-60 seconds in environment setup time during CI as well, though that isn't really a big deal for an hour and a half test suite :) 
    * The only issue I came across was the test_mdf issue for old versions of onnxruntime. The issue results from the fact that pip pre-compiles bytecode when packages are installed and uv doesn't by default (though this can be [enabled](https://docs.astral.sh/uv/pip/compatibility/#bytecode-compilation)). The old versions of onnxruntime don't cause the crash in pytest when onnxruntime is pip installed because they are already compiled at pytest import time and don't trigger the SyntaxWarning from the invalid escape sequence. I could enable the pre-compiled bytecode to get around this but I don't see a lot of folks running uv with that option so I wanted to keep things more standard. We have dropped python 3.8 (and probably will drop 3.9 soon) on the MDF package so I am not too concerned about skipping the test in this case.
    * The exclude of broken Python 3.9 on ARM seems to be the same issue that was causing us to exclude 3.8. I just confirmed it was broken on 3.9 as well so I moved the exclude from 3.8 to 3.9. I think its because it is defaulting to ARM64 and setup python action doesn't support <3.10 on ARM64 MacOS. Once again, I don't think we should be spending time trying to support end-of-life python versions for this project. I think if the setup-python action doesn't support the combination we are good to drop it as well.

* **`setup-uv` action**:  Agreed we should be careful with third‑party actions. I did a quick review, it’s the official action from the uv maintainers (Astral), has a small surface area (download + cache + PATH), and publishes tagged releases. If you’d prefer zero third‑party actions, I can switch to: `actions/setup-python@v5` + `python -m pip install 'uv==<pinned_version>'` and proceed with uv from there. Or maybe add this as a fallback option? I think the risk of abandonment is small because we are still using the pip interface, it wouldn't be hard to switch back to pip.

kmantel:
I agree with @jvesely on splitting it up

I also think it's good to keep support for python 3.8 (/older versions in general) as long as it's not causing too much of a problem and we aren't specifically looking for new features.

I'm not totally convinced on the benefits of lockfile/local-venv like used in uv. If we were making helper tools for students having trouble creating environments, I could see it having a place there, but I also think conda is still more of a known factor there. I also have some reservations about depending on uv. It's still pretty new and managed by Astral, a VC-backed company without current monetization. I'm concerned about what it looks like in the near future when they've gained enough user-share and decide it's time to make money. I suppose this also applies to ruff, but if that supports additional checks that we can't do with flake8 or code/docstyle, it's less of a commitment.

As for packaging, when I was updating it to use pypa/build, maybe I ran into some of the same secondary effects on pyprojection.toml, which made me not end up continuing that way. There were definitely some things to work out around the specific contents of the wheels and source distributions; it looks like at least the setuptools-scm replacement of versioneer may affect this (https://github.com/jwodder/versioningit/issues/46#issuecomment-1501201632). I'd also like to know that https://github.com/PrincetonUniversity/PsyNeuLink/actions/workflows/test-release.yml would still work, because that was pretty finicky as well.

davidt0x:
> I agree with @jvesely on splitting it up

Ok, so I think I will close the PR and try to split it up. I think I will just make a more focused PR for moving to pyproject.toml. I think we should move off versioneer since it hasn't been updated since 2023 and still requires setup.py. I picked setuptools_scm since its what I knew but I am open to other things. I believe it still respects MANIFEST.in for sdists so I kept it. By default it includes everything in SCM into the sdist I think.

> I also think it's good to keep support for python 3.8 (/older versions in general) as long as it's not causing too much of a problem and we aren't specifically looking for new features.

Ok, I guess we can postpone this.

> I'm not totally convinced on the benefits of lockfile/local-venv like used in uv. If we were making helper tools for students having trouble creating environments, I could see it having a place there, but I also think conda is still more of a known factor there. I also have some reservations about depending on uv. It's still pretty new and managed by Astral, a VC-backed company without current monetization. I'm concerned about what it looks like in the near future when they've gained enough user-share and decide it's time to make money. 

Yeah, I hear the concerns about Astral and their eventual need to make a profit. I thought about it a lot myself and a couple of things have made me less concerned. First, it is open source so if they do decide to start charging for things it will almost certainly be forked. See what has happened with conda for example, conda-forge has kind of taken over at this point it seems. That was an even more difficult situation since it is a package index. Second, they have built uv to be in line with the python packaging standards. Since we are still using the pip interface, migrating back to pip would basically amount to removing the uv command from `uv pip`. The great thing about uv over conda is that it supports the standard venv\pip workflow. It is also makes setting up python on all platforms really easy because it downloads standalone relocatable python versions for you. In my opinion, it is the best piece of python tooling in the last five years. My view is that Astral will probably eventually start charging corporate clients to make their money, similar to Anaconda. That being said, I think we can hold off at least on moving to it in CI, since we have very long runs anyway the benefits of its speed are kind of negligible.

> I suppose this also applies to ruff, but if that supports additional checks that we can't do with flake8 or code/docstyle, it's less of a commitment.

I actually didn't notice this until now but [pydocstyle](https://github.com/PyCQA/pydocstyle) is actually deprecated. They recommend moving to ruff. pycodestyle is still maintained, but it seems they [don't want](https://github.com/PyCQA/pycodestyle/issues/1226) to support pyproject.toml. So we have to move on from pydocstyle at least, I predict pycodestlye will eventually be deprecated as well. May as well just move now to the better tool that does both things (and much more).

> I'd also like to know that https://github.com/PrincetonUniversity/PsyNeuLink/actions/workflows/test-release.yml would still work, because that was pretty finicky as well.

Yeah, this stuff could definitely break the release workflow. At first glance I don't see anything in there that should break but there might be something I am missing. Not sure of the best way to test it though without merging things and just trying it. I am sure we can work out the issues that come up.

jvesely:
pyproject.toml
agreed. Looks like a good idea. I still think doing it in a separate PR is better/more readable, even if there are prerequisites on some of the other changes.

ruff:
I don't have a strong preference for using a specific linting tool. The need to add extra noqa seemed like a step backward, hence the question. As long as the general direction of increasing enforcement of consistent coding style rules is maintained, a PR switching to ruff should be OK and fairly straightforward. Agreed with getting rid of setup.cfg after toml migration.
sounds like this should be the first PR in the series.

Python 3.8:
I don't have a strong preference about keeping it. I don't mind having a predictable "we drop Python X N months after EOL" policy, although the current "we drop Python when it becomes too cumbersome to maintain" has worked ok.
I don't think PNL has taken the full advantage of having even Python 3.8 features yet (e.g. functools.cached_property).

The sequence of commits looks like it was dropped to workaround the issue with onnxruntime, which would be a concern (see below).

uv migration:
I think this needs more discussion.
One aspect is the direction and whether we want to add dependency on uv in place of the older existing tools. We can be pretty sure that pip and co. will be around and supported in some way for years to come. I'm not sure we can say the same thing about uv.
The main question is whether this would impact the way users install PNL. I'd guess most use pip install or conda. If a switch to uv mean we're no longer testing vanilla pip install, and there are differences in behaviour, we won't catch bugs there.
If we decide to use uv then using the action setup-uv should be fine as well, since it's the same maintainer.

The other aspect is technical.
I'm concerned about the onnxruntime issue that led to putting a lower bound and dropping Python 3.8. I suspect that it might be related to mixing uv package handling with setup-python.
The issue with onnxruntime in general is concerning because it suggests that uv installation behaves differently from vanilla pip install, which is not what we want in CI testing.
If we want to switch to uv, it might work better to use it to manage Python versions as well and drop setup-python.

Another concern here is how this handles packages built from sdist. Currently, that's only pycodestyle/pycdostyle, which would be going away with ruff. However, at times, it included building onnx from source. This should all work, and the existing pip cache maintenance takes extra care not to clean up locally built versions of the most recent dependencies while at the same time removing all downloaded wheels. I'm not sure `uv cache purge` matches that behaviour.


> Not sure of the best way to test it though without merging things

You can create/push tags in a forked repo and that should run most of the release scripts (no uploading to pypi)

PR #3392, opened 2025-11-12
llvm: Improve UDF compiler

jvesely:
Add UDF compiler support for explicit Python type casts (float, int, bool).
Add UDF compiler support for rugged list literals.
Add UDF compiler support for constant indices to rugged lists.
Add UDF compiler support for "IfExpression" / ternary operator. The support is limited to cases in which both branches return values of the same type.
Add UDF compiler support for 'np.argmax'.
Add UDF compiler support for 'np.nan'.

### DIFF
--- a/psyneulink/core/components/mechanisms/mechanism.py
+++ b/psyneulink/core/components/mechanisms/mechanism.py
@@ -3154,8 +3154,6 @@ class Mechanism_Base(Mechanism):
         else:
             aggregate_type = pnlvm.ir.LiteralStructType(types)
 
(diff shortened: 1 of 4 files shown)


jdcpni:
This looks pretty cool!
jdc

Sent from my iPhone

On Nov 12, 2025, at 7:24 PM, Jan Vesely ***@***.***> wrote:

﻿

Merged #3392<https://github.com/PrincetonUniversity/PsyNeuLink/pull/3392> into devel.

PR #3396, opened 2025-11-14
support rich v11+

kmantel:


jvesely:
Should we just bump minimum rich version to >=11 and switch to 'Group'?
11.0.0 was released in Jan 2022, so it should be good for Python 3.8+.

kmantel:
It is just a simple rename with no interface changes, but looking further, they did the rename in 10.7:

https://github.com/Textualize/rich/blob/4d6d631a3d2deddf8405522d4b8c976a6d35726c/CHANGELOG.md?plain=1#L719-L733

How about just bumping to there?

PR #3430, opened 2026-01-03
tests: Use pattern matching to check for expected warnings and errors

jvesely:
Enable compiled variant of the full `test_transfer_mech_integration_rate_0_8_initial_0_5` test, including direct modification of the `noise.base` value.
Use error action on warnings instead of checking the number of warnings collected via `catch_warnings`.
Use regexp, or regexp escaped string, in `match` parameter to check for expected warnings.
Set minimum pytest version to >8.0.2 to allow nested pytest.warns() constructs.


jvesely:
hi @kmantel,
do you mind taking a look if this breaks anything major?
The main goal is to "unhide" warnings that were collected in unrestricted `pytest.warns` contexts. The rest is mostly refactoring/consolidation of tests around the above.
The only non-test change is adding 'name' parameter to AdaptiveIntegrator.

There is still a Component naming issue for components that are instantiated during pnl module load (and the tests that were switched to explicit names of AdaptiveIntegrator functions hit that), but that's not addressed in this change, and the use of explicit names is just a workaround.

EDIT: The workarounds for AdaptiveIntegrator names can be removed after https://github.com/PrincetonUniversity/PsyNeuLink/pull/3431 lands

kmantel:
Yeah, I agree that using the match arg for warns/raises is the way to go. I actually made a similar change to `test_learning_rate_specification_errors` on a working branch. I also see that the assert in `test_learning_rate_with_3_levels_of_nesting_and_build_pytorch_rep_spec` wasn't doing what was intended, so that's good.

There's one odd thing - I see the tests are passing, but I checked `test_add_node_with_controller_spec_and_control_mech_but_not_a_controller` because I wasn't familiar with doing this:

https://github.com/PrincetonUniversity/PsyNeuLink/blob/9b987cb85a41a2f01c8f2078206d84dfc1d51ab8/tests/composition/test_control.py#L2412-L2414

This fails for me with
```
E            Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted.
E            The list of emitted warnings is: [].
```

using python 3.11 and pytest 7.4.3. Apparently, using multiple `pytest.warns` is only supported in newer pytest versions (https://github.com/pytest-dev/pytest/issues/12456 says 8+ but I didn't confirm), so we'd have to bump the pytest floor.

~I also didn't catch why the parametrizations in `tests/mechanism/test_ddm_mechanism.py` were removed.~ (Was checking against current devel rather than base)

jvesely:
> using python 3.11 and pytest 7.4.3. Apparently, using multiple `pytest.warns` is only supported in newer pytest versions ([pytest-dev/pytest#12456](https://github.com/pytest-dev/pytest/issues/12456) says 8+ but I didn't confirm), so we'd have to bump the pytest floor.

Thanks for catching this. Issue [pytest-dev/pytest#12456](https://github.com/pytest-dev/pytest/issues/12456) is marked as "closed as not planned", so my guess is that it's https://github.com/pytest-dev/pytest/pull/10937 that enabled nesting. Either way. I've added pytest>=8.0.2 as a lower bound

> I also didn't catch why the parametrizations in tests/mechanism/test_ddm_mechanism.py were removed. (Was checking against current devel rather than base)

Yeah, I have just rebased the PR. It might be cleaner to get #3431 merged first so that the naming changes (and enabling names for AdaptiveIntegrator) can be moved to a separate PR.

PR #3431, opened 2026-01-04
Registries: Use '_pnl_internal_' prefix for Components created during module load

jvesely:
Assert that internal Components are not compiled.

### DIFF
--- a/psyneulink/__init__.py
+++ b/psyneulink/__init__.py
@@ -99,6 +99,8 @@ for reg in primary_registries:
 
     process_registry_object_instances(reg, func)
 
+core.globals.registry._register_auto_name_prefix = ""
+
 def System(*args, **kwars):
     show_warning_sys_and_proc_warning()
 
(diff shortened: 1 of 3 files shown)


kmantel (commented):
Sorry I missed this request.

I think this is fine. 

I wonder if we should rename any of `_PNL_INTERNAL_PREFIX = "__pnl_internal_"`, `Component.is_pnl_inherent`, and `Parameter.pnl_internal`. The last one was meant to label parameters that are psyneulink-specific and not part of the MDF spec or used to read/write an MDF model. I don't have a really strong opinion on this.

jvesely:
> Sorry I missed this request.
> 
> I think this is fine.
> 
> I wonder if we should rename any of `_PNL_INTERNAL_PREFIX = "__pnl_internal_"`, `Component.is_pnl_inherent`, and `Parameter.pnl_internal`. The last one was meant to label parameters that are psyneulink-specific and not part of the MDF spec or used to read/write an MDF model. I don't have a really strong opinion on this.

Thank you. I think it makes sense to keep the naming consistent. I've changed the names to "inherent". If we ever decide that there's a better name, at least it will be clear that both places should be changed.

PR #3432, opened 2026-01-06
ci: Remove .python-version

jvesely:
It was a necessary workaround for Dependabot using lowest compatible Python version when scanning for dependency updates. This behaviour was changed to prefer the highest compatible Python version in [1], and the workaround is no longer necessary.

[1] https://github.com/dependabot/dependabot-core/pull/13862


jvesely:
Dependabot now uses Python 3.14 for PNL[1]:

```
updater | 2026/01/06 02:01:12 INFO <job_1200271997> Detected python version: 3.14.2
```
[1] https://github.com/jvesely/PsyNeuLink/actions/runs/20735388109/job/59531591427

PR #3460, opened 2026-02-07
mdf: Flush file contents when using temporary files

jvesely:
Otherwise the written contents might not be visible to mdf parser.

Fixes `tests/mdf/test_mdf.py::test_generate_script_from_mdf[yml-model_basic.py-comp]` test failure on Python 3.14


jvesely:
cc @kmantel

PR #3481, opened 2026-03-18
TransferMechanism: fix non-default integrator_function shape mismatch

kmantel:
TransferMechanism._parse_function_variable bypassed the integrator_function, which would fail to update the variable shape of TransferMechanism.function

This commit causes test result differences in
tests/mechanisms/test_transfer_mechanism.py::TestDistributionFunctions and
tests/documentation/test_module_docs.py::test_core_docs[psyneulink.core.components.mechanisms.processing.transfermechanism] due to additional calls to noise functions during TransferMechanism initialization causing RNG state offset. The new results can be replicated on the parent commit by manually calling the noise functions in TransferMechanism._parse_function_variable:

```
     def _parse_function_variable(self, variable, context=None):
         if self.is_initializing:
+            self._try_execute_param(self.parameters.noise._get(context), variable, context)
             return super(TransferMechanism, self)._parse_function_variable(variable=variable, context=context)
```


davidt0x (approved):
LGTM

PR #3498, opened 2026-04-10
Component: Add support for modulating nested Components

jvesely:
Try to use the port associated with modulated Parameter directly.
Add test.


kmantel (approved):
I think this is fine.


I think `parameter.final_source.port` should end up being the same object we want from `parameter_port_list` anyway, but yeah that's tangential.

PR #3527, opened 2026-04-28
llvm, TransferMechanism: Fix compiled reset semantics to match Python in integrator mode

jvesely:
Allow any Sequence for tags when selecting the tested Function and Mechanism variant.
Remove the 'reset' tag, and use the integrator function output when running the main Function in the mechanism reset.

PR #3530, opened 2026-04-30
Refactor LCA mechanism output ports and improve PEC

davidt0x:
## Summary

This PR is the cleaned-up follow-on to the earlier PEC/LCA branch.

The generic mechanism-wide output-port execution count fix was split out and merged separately in #3529, so this PR now focuses on:

- adding the retained LCA standard output ports
- fixing PEC outcome-variable handling for multi-output / LCA models
- keeping the minimal LLVM/runtime fixes still required for the new PEC/LCA coverage on current `devel`

In particular, this PR keeps the PEC- and LCA-related work from the original branch, but does not reintroduce the broader compilation/LLVM patch stack that had accumulated there.

### 1. Add LCA standard output ports

`LCAMechanism` now exposes three additional standard output ports:

- `DECISION_INDEX`
- `DECISION_STEPS`
- `DECISION_TIME`

These are exported as `pnl.DECISION_*`, added to `LCAMechanism.standard_output_ports`, and added to `standard_output_port_names`.

Their semantics are:

- `DECISION_INDEX`: index of the winning element in the final LCA value
- `DECISION_STEPS`: final `num_executions_before_finished`
- `DECISION_TIME`: `DECISION_STEPS * time_step_size`

This makes decision metadata directly available through the standard output-port interface already used by compositions and downstream analyses.

### 2. Fix PEC outcome-variable indexing for multi-output models

`ParameterEstimationComposition` previously treated outcome-variable indices as though each output port contributed a single scalar slot.

That breaks for models that expose multiple output ports and for ports whose values flatten into multiple columns, including the new LCA decision outputs.

This PR fixes that by flattening outcome-variable indices across output-port widths, so PEC selects the correct columns from composition results even when:

- a mechanism exposes multiple output ports
- an output port contributes more than one value
- the target outcome comes from an LCA decision-related output port

### 3. Make PEC/LCA regression coverage deterministic

The PEC tests now use explicit ordered pathway lists instead of relying on unordered set-based pathway specifications.

That keeps input/output port ordering stable and avoids spurious failures unrelated to the actual PEC logic being tested.

### 4. Keep the minimal LLVM/runtime fixes required by the retained PEC/LCA path

A small number of focused runtime/compiler fixes are still needed on current `devel` for these PEC/LCA cases:

- `optimizationfunctions.py`
  - use `math.prod(...)` for compiled `num_evals`
  - this avoids `np.prod([]) -> 1.0` leaking a float into shape/count logic for empty search spaces

- `distributionfunctions.py`
  - make `NormalDist`'s LLVM path handle singleton-array outputs
  - this fixes the `[1 x double]*` vs `double*` mismatch that appears when `NormalDist` is used as LCA noise in compiled execution

- `parameters.py`
  - handle compiled-tracked boolean parameter updates correctly
  - this is needed when a compiled object flips a boolean parameter after compilation, rather than only at construction time

  In particular, this showed up in
  `tests/mechanisms/test_lca.py::TestLCA::test_LCAMechanism_standard_output_ports[ExecutionMode.LLVMRun]`,
  where the test runs an `LCAMechanism` with `execute_until_finished=True`, then flips
  `execute_until_finished` to `False` and runs again in `LLVMRun`.

  Without this change, the compiled parameter update can be incorrectly treated as a type-changing scalar update rather than an in-place update to the compiled struct. In that case, LLVM continues behaving as though `execute_until_finished` were still `True`.

  The symptom in the regression test is that after switching `execute_until_finished=False`, the next LLVM run should report `DECISION_STEPS == 1`, but instead it continues the prior until-finished behavior and reports the previous full decision count.

### 5. Align `RecurrentTransferMechanism` reset semantics with the generic reset fixes already merged in `devel`

This PR also keeps a `RecurrentTransferMechanism` LLVM reset fix that is still needed on top of current `devel`.

This is closely related to #3527, which fixed compiled reset semantics for the generic `Mechanism` / `TransferMechanism` integrator-mode path. In #3527, the reset path was updated so that compiled execution matches Python by:

- resetting the integrator function
- feeding the reset integrator output through the main function
- updating output ports from that reinitialized value

However, `RecurrentTransferMechanism` overrides `_gen_llvm_function_reset()`, so it does not automatically inherit the full behavior fixed in #3527.

That matters here because `LCAMechanism` is a subclass of `RecurrentTransferMechanism`, and its recurrent feedback in LLVM is seeded from RTM state (`old_val`). Without the RTM-specific reset fix, the recurrent state after reset is not refreshed from the reset output-port values, which causes LLVM/Python divergence across trials.

In practice, this shows up in the LCA standard-output-port regression:
- Python reports the expected post-reset decision-step count
- LLVM finishes one step early after reset unless the RTM reset path is aligned

So this RTM change should be viewed as the subclass-specific follow-up to #3527, not as a separate broad compiler change.

## Why

The goal here is to make LCA decision metadata directly consumable by users and by PEC, while keeping the branch narrowly scoped.

After splitting the generic mechanism execution-count fix into #3529, the remaining work for this PR is:

- expose the LCA outputs themselves
- make PEC read those outputs correctly
- keep the minimal supporting runtime behavior needed for those paths to work on current `devel`

## Testing

Targeted validation for this PR includes:

- PEC input-format regressions
- PEC outcome-variable selection for multi-output / LCA models
- LCA standard output port coverage
- compiled PEC/LCA regressions involving stochastic LCA noise
- LLVM reset behavior needed for LCA/RTM consistency across trials

Relevant tests include:

- `tests/composition/pec/test_parameterestimationcomposition.py::test_pec_run_input_formats[pec_good]`
- `tests/composition/pec/test_parameterestimationcomposition.py::test_pec_run_input_formats[model_good]`
- `tests/composition/pec/test_parameterestimationcomposition.py::test_pec_lca`
- `tests/composition/pec/test_parameterestimationcomposition.py::test_pec_lca_num_estimates_llvm_stochasticity`
- `tests/mechanisms/test_lca.py::TestLCA::test_LCAMechanism_standard_output_ports`
- `tests/mechanisms/test_recurrent_transfer_mechanism.py::TestRecurrentTransferMechanismReset`

## Notes

- This PR assumes #3529 is already in `devel` and intentionally does not duplicate that mechanism-wide fix.
- The `RecurrentTransferMechanism` reset change is intentionally retained even though #3527 is already merged, because RTM overrides the reset path and still needs the subclass-specific alignment described above.


psyneulink/library/components/mechanisms/processing/transfer/recurrenttransfermechanism.py
```diff
@@ -1291,8 +1291,27 @@ def _gen_llvm_function_reset(self, ctx, builder, params, state, arg_in, arg_out,
...
+            f_out = builder.alloca(f.args[3].type.pointee, name="function_reinit_out")
+            builder.call(f, [f_params, f_state, reinit_out, f_out])
+            reinit_out = f_out
+
+        # Keep recurrent feedback state aligned with reset output-port values.
+        # Python reset updates output ports, and recurrent input on the next
+        # step is based on those refreshed outputs.
+        builder = self._gen_llvm_output_ports(
+            ctx, builder, reinit_out, params, state, arg_in, arg_out
+        )
```
jvesely: This change is wrong, but there's a simpler solution I'll put up a PR shortly.

jvesely: This part should no longer be necessary after #3532

psyneulink/core/components/functions/nonstateful/distributionfunctions.py
```diff
@@ -214,12 +214,15 @@ def _gen_llvm_function_body(self, ctx, builder, params, state, _, arg_out, *, ta
...
-        std_dev = builder.load(std_dev_ptr)
+        mean = pnlvm.helpers.load_extract_scalar_array_one(builder, mean_ptr)
+        std_dev = pnlvm.helpers.load_extract_scalar_array_one(builder, std_dev_ptr)
 
         ret_val = builder.fmul(ret_val, std_dev)
         ret_val = builder.fadd(ret_val, mean)
 
+        while isinstance(arg_out.type.pointee, pnlvm.ir.ArrayType):
+            assert len(arg_out.type.pointee) == 1
+            arg_out = builder.gep(arg_out, [ctx.int32_ty(0), ctx.int32_ty(0)])
```
jvesely, line 228: Unlimited indexing should never be needed. If there's a specific dimensionality issue, it should be addressed in a more targeted way.

psyneulink/core/globals/parameters.py
```diff
@@ -1870,8 +1871,15 @@ def _set_value(
 
         value_updated = False
         if not compilation_sync:
+            value_for_update = value
+
+            # bools are stored as ints in compiled structs; convert here to
+            # avoid incorrectly flagging type change
```
jvesely, line 1877: Why would this flag a type change? Updating int arrays with bools is ok.
```
>>> a = np.array([1,2,3])
>>> a[1] = False
>>> a
array([1, 0, 3])
...
>>> b = np.array(5)
>>> b[...] = False
>>> b
array(0)
```

psyneulink/library/components/mechanisms/processing/transfer/lcamechanism.py
```diff
@@ -202,16 +202,20 @@
...
 from psyneulink.core.globals.preferences.basepreferenceset import ValidPrefSet
 from psyneulink.library.components.mechanisms.processing.transfer.recurrenttransfermechanism import \
     RecurrentTransferMechanism, _recurrent_transfer_mechanism_matrix_getter, _recurrent_transfer_mechanism_matrix_setter
 
-__all__ = ['LCAMechanism', 'LCAError', 'CONVERGENCE']
+__all__ = ['LCAMechanism', 'LCAError', 'DECISION_INDEX', 'DECISION_STEPS', 'DECISION_TIME']
 
+DECISION_INDEX = 'DECISION_INDEX'
+DECISION_STEPS = 'DECISION_STEPS'
+DECISION_TIME = 'DECISION_TIME'
```
jvesely, line 218: Should these be in `keywords`?

tests/mechanisms/test_lca.py
```diff
@@ -277,6 +277,68 @@ def test_LCAMechanism_termination_threshold(self):
...
+            execute_until_finished=True,
+            reset_stateful_function_when=pnl.AtTrialStart(),
+            output_ports = [pnl.RESULT,
+                            pnl.DECISION_INDEX,
+                            pnl.DECISION_STEPS,
+                            pnl.DECISION_TIME
+                            ]
+        )
+        if comp_mode not in {pnl.ExecutionMode.Python, pnl.ExecutionMode.LLVMRun}:
+            pytest.skip("LCAMechanism_standard_output_ports only works for ExecutionMode = Python or LLVMRun.")
```
jvesely: It should also work in LLVMExec and PTXRun. Please don't skip those tests.

tests/mechanisms/test_lca.py
```diff
@@ -277,6 +277,68 @@ def test_LCAMechanism_termination_threshold(self):
...
+        comp = pnl.Composition(lca)
+        lca.execute_until_finished = True
+
+        def assert_expected_results(results, expected):
```
jvesely: This will break error reporting because the assertion failures are on the same line for all tested outcomes.

tests/mechanisms/test_lca.py
```diff
@@ -277,6 +277,68 @@ def test_LCAMechanism_termination_threshold(self):
...
+            index_val, steps_val, time_val = expected
+            if comp_mode is pnl.ExecutionMode.Python:
+                assert lca.output_ports[pnl.DECISION_INDEX].value == index_val
+                assert lca.output_ports[pnl.DECISION_STEPS].value == steps_val
+                assert lca.output_ports[pnl.DECISION_TIME].value == time_val
```
jvesely: Comparisons of floating-point values should be done using `allclose`.

jvesely (changes requested):

tests/mechanisms/test_lca.py
davidt0x: Yeah, these seem to pass, not sure why Jon had it in there.

psyneulink/core/components/functions/nonstateful/distributionfunctions.py
davidt0x: Yes, I tightened this up.

I replaced the open-ended loop with a single, type-checked unwrap that only handles the specific case we’re actually seeing here: a scalar `ret_val` being stored into an `arg_out` with pointee type `[1 x double]` rather than `double`. So now we only peel one singleton array layer, and we assert that the element type matches the scalar being stored.

I also checked whether `pnlvm.helpers.unwrap_2d_array()` would cover this, but it doesn’t in this case. That helper only unwraps pointers whose pointee is an array of arrays, whereas the failing `NormalDist` / `UniformDist` case here is a singleton 1D array pointer (`[1 x double]*`). Using `unwrap_2d_array()` left `arg_out` unchanged and reproduced the original LLVM type mismatch.

psyneulink/core/globals/parameters.py
davidt0x: Katherine had a look at this change and made some fixes. I think there may be some remaining underlying issues between synchronizing Python and compiled side types in the case when a user runs a composition and then changes a parameter to another type. This is just a targeted fix for the boolean case.

psyneulink/library/components/mechanisms/processing/transfer/lcamechanism.py
davidt0x: I think these should stay local to `lcamechanism.py`.

The closest precedent is `DDM`, which defines/export its mechanism-specific standard output port names (`DECISION_VARIABLE`, `RESPONSE_TIME`, `DECISION_OUTCOME`, etc.) in `ddm.py` rather than `keywords.py`. Since `DECISION_INDEX`, `DECISION_STEPS`, and `DECISION_TIME` are likewise LCA-specific standard output port names rather than generic framework keywords, keeping them alongside `LCAMechanism` seems more consistent with the existing pattern. 

I do think they should probably be unified with DDM though at some point. But I don't think I want to do that in this PR.

tests/mechanisms/test_lca.py
davidt0x: Agreed. I’ll inline those assertions so pytest points directly at the failing outcome instead of routing everything through the helper.

tests/mechanisms/test_lca.py
davidt0x: Good point. I changed the test so the floating-point `DECISION_TIME` checks now use `np.testing.assert_allclose`, while `DECISION_INDEX` and `DECISION_STEPS` still use exact equality.

psyneulink/core/components/functions/nonstateful/distributionfunctions.py
davidt0x: I tightened this further.

The original loop is gone, but it turned out there are two concrete wrapped output shapes showing up in compiled call sites:

- a 2D mechanism/control case (`[1 x [1 x double]]*`)
- a remaining singleton-array scalar case (`[1 x double]*`)

So the updated fix now:

1. uses `unwrap_2d_array()` for the explicit 2D wrapper case, and then
2. does one type-checked single-level unwrap only if the scalar still does not match the pointee type.

That keeps the behavior targeted to the known shapes we actually hit here, rather than peeling arbitrary nesting.

tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -559,3 +581,138 @@ def test_pec_controller_specified():
...
+    input_dict = {
+        input_mech: np.tile([1,0], (100,1))
+    }
+
+    # Generate data to fit
+    execution_mode_map = {
+        "Python": pnl.ExecutionMode.Python,
+        "LLVM": pnl.ExecutionMode.LLVMRun,
+        "PTX": pnl.ExecutionMode.PTXRun,
+    }
```
jvesely, line 628: Not for this PR, but would there be enough uses for "pec_mode" fixture that would iterate over `ExecutionMode.Python, ExecutionMode.LLVMRun, ExecutionMode.PTXRun`? maybe even without the Python part?

tests/mechanisms/test_lca.py
```diff
@@ -277,6 +277,85 @@ def test_LCAMechanism_termination_threshold(self):
...
+        comp = pnl.Composition(lca)
+        lca.execute_until_finished = True
+
+        lca.parameters.time_step_size.set(.01, comp.name)
+        actual = comp.run(inputs=[[1, 0]], execution_mode=comp_mode)
+        if comp_mode is pnl.ExecutionMode.Python:
+            np.testing.assert_equal(lca.output_ports[pnl.DECISION_INDEX].value, 0)
+            np.testing.assert_equal(lca.output_ports[pnl.DECISION_STEPS].value, 14)
+            np.testing.assert_allclose(lca.output_ports[pnl.DECISION_TIME].value, .14)
+        else:
```
jvesely: Does this need to be in "else" branch? Do Python results differ from the compiled ones?

tests/mechanisms/test_lca.py
```diff
@@ -277,6 +277,85 @@ def test_LCAMechanism_termination_threshold(self):
...
+                            pnl.DECISION_STEPS,
+                            pnl.DECISION_TIME
+                            ]
+        )
+        comp = pnl.Composition(lca)
+        lca.execute_until_finished = True
+
+        lca.parameters.time_step_size.set(.01, comp.name)
+        actual = comp.run(inputs=[[1, 0]], execution_mode=comp_mode)
+        if comp_mode is pnl.ExecutionMode.Python:
```
jvesely: yeah, "value" parameter is only enabled on Mechanisms in compiled mode. Would it be useful outside testing to enable it for output ports as well?

psyneulink/core/components/functions/nonstateful/distributionfunctions.py
jvesely: Thanks for doing this.
There are still places that have unfortunate constructs all over the compiled code. It helps to clean them up when encountered, so they don't proliferate.

psyneulink/core/globals/parameters.py
jvesely: OK. thank you. I'll leave this to you and @kmantel  then.
The original idea was that assignments can just do:
```
   try:
      old_val[...] = new_val
   except ValueError:
      # need to flush compiled state here
```
but if there's a different approach, it's up to the Parameter class.

tests/mechanisms/test_lca.py
jvesely: Can the checks be combined into a single line?
```
np.assert_allclose(result, expected)
```

jvesely (approved):

jvesely:
just a few optional nits and questions

psyneulink/core/globals/parameters.py
kmantel: @jvesely there's a problem with just trying to assign directly like that, where if the assignment fails, the array may still be partially updated:

```
>>> import numpy as np
>>> a = np.array([0, 1])
>>> a[...] = [True, 'invalid type']
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    a[...] = [True, 'invalid type']
    ~^^^^^
ValueError: invalid literal for int() with base 10: 'invalid type'
>>> a
array([1, 1])
```

I can't remember exactly when this came up in practice, but if I had to guess maybe in a noise spec with mixed numbers/`Function`s.

It was also my impression that something like `a[...] = [0.1, 1.1]` should mean a type change and recompile, because otherwise, the floats are truncated.

So to deal with this, `update_array_in_place` does some type/shape/broadcast checks before using `np.copyto`, under the assumption that anything compatible would already be a numpy array by the time it was called. It might be possible just to rely on the `casting` arg, but I'm not sure. Might be an issue with [potentially nested] ragged arrays.

jvesely: Thanks. Makes sense, the partial update behaviour makes that approach a lot less useful than I assumed.

> a[...] = [0.1, 1.1] should mean a type change and recompile

testing locally on 3.13, this truncates the new values rather than converting the type:

```
>>> a = np.array([1,2,3])
>>> a.dtype
dtype('int64')
>>> a[...] = [1.1, 2.2, 3.3]
>>> a
array([1, 2, 3])
>>> a.dtype
dtype('int64')
```

Silent information loss is probably not what we want either, but the situation is more theoretical.
All compiled numeric values are floats (float64 by default), except for a few flag and count Parameters.

Theoretically, it can happen in the other direction; if the default is fp32 (e.g. `pytest --fp-precision=fp32`) and we try to assign int64.

tests/mechanisms/test_lca.py
davidt0x: Ok, yeah, I have cleaned this up, didn't need the else.

tests/mechanisms/test_lca.py
davidt0x: Yeah, not sure if its useful outside testing.

tests/composition/pec/test_parameterestimationcomposition.py
davidt0x: Not sure, honestly, for most of the PEC tests, skipping the Python execution is needed. I have a lot of tests were I skip if because of speed.

PR #3533, opened 2026-05-06
LinearCombination: higher-dimension capable LLVM implementation

kmantel:
supports parameter specifications as scalar, len-1 vector, or N-dim array matching shape of input (weights, exponents) or output (scale, offset)

(len-M vector elementwise for (M, ...) shape arrays is intended to work, but does not work in non-compiled PNL)

### DIFF
--- a/psyneulink/core/components/functions/nonstateful/transformfunctions.py
+++ b/psyneulink/core/components/functions/nonstateful/transformfunctions.py
@@ -112,8 +112,13 @@ class TransformFunction(Function_Base):
             return ctx.float_ty(default)
 
         elif isinstance(param_type, pnlvm.ir.ArrayType):
-            index = ctx.int32_ty(0) if len(param_type) == 1 else index
-            param_ptr = builder.gep(param_ptr, [ctx.int32_ty(0), index])
+            if len(param_type) == 1 and not isinstance(param_type.element, pnlvm.ir.ArrayType):
+                index = [ctx.int32_ty(0)]
+
+            if not isinstance(index, list):
+                index = [index]
+
+            param_ptr = builder.gep(param_ptr, [ctx.int32_ty(0), *index])
 
         return builder.load(param_ptr)
 
@@ -1533,10 +1538,92 @@ class LinearCombination(
         default_var = np.atleast_2d(self.defaults.variable)
         return ctx.convert_python_struct_to_llvm_ir(default_var)
 
-    def _gen_llvm_combine(self, builder, index, ctx, vi, vo, params):
-        scale = self._gen_llvm_load_param(ctx, builder, params, SCALE, index, 1.0)
-        offset = self._gen_llvm_load_param(ctx, builder, params, OFFSET, index, -0.0)
+    def _gen_llvm_function_body(self, ctx, builder, params, _, arg_in, arg_out, *, tags: frozenset):
+        # for the frequent case where non-LLVM variable is 2D and value is cast
+        # up to 2D from 1D. arg_out is expected to have one fewer dimension than
+        # arg_in in _gen_llvm_combine_body.
+        if (arg_in.type.pointee.element == arg_out.type.pointee.element) and arg_out.type.pointee.count == 1:
+            arg_out = pnlvm.helpers.unwrap_2d_array(builder, arg_out)
 
+        self._gen_llvm_combine(builder, ctx=ctx, vi=arg_in, vo=arg_out, params=params)
+        return builder
+
+    def _gen_llvm_combine_body(self, builder, ctx, vi, vo, val, pow_f, comb_op, params, vo_idx):
+        ptro = builder.gep(vo, [ctx.int32_ty(0), *vo_idx])
+
+        if isinstance(ptro.type.pointee, pnlvm.ir.ArrayType):
+            # Recursion on output array dimensions stops when pointing to scalar
+            with pnlvm.helpers.array_ptr_loop(builder, ptro, f"combine_axis{len(vo_idx)}") as (b, idx):
+                self._gen_llvm_combine_body(b, ctx, vi, vo, val, pow_f, comb_op, params, [*vo_idx, idx])
+        else:
+            # val is the base value passed in according to the combination
+            # operation (0 for sum, 1 for product)
+            val_p = builder.alloca(val.type, name="combined_result")
+            builder.store(val, val_p)
+
+            assert isinstance(vi.type.pointee, pnlvm.ir.ArrayType)
+            with pnlvm.helpers.array_ptr_loop(builder, vi, "combine") as (b, idx):
+                # The recursion is on the output, so `vo_idx` corresponds to a
+                # scalar in the output array. This loop is over the input, which
+                # has shape (vi.type.pointee.count, <`vo` shape>) and assumes axis=0.
+                # So, we combine elements [0, `vo_idx`]...[`vi.type.pointee.count-1`, `vo_idx`]
+                # to get the result element (`vo_idx`) in the output.
+                in_idx = [idx, *vo_idx]
+                ptri = b.gep(vi, [ctx.int32_ty(0), *in_idx])
+                in_val = b.load(ptri)
+
+                # lower-dim array specification (ex: exponents [a, b]
+                # and var [[[1, 1]], [[2, 2]]], applying a to [[1, 1]]
+                # and b to [[2, 2]])
+                exponent = self._gen_llvm_load_param(ctx, b, params, EXPONENTS, idx, 1.0)
+                if (
+                    isinstance(exponent.type, pnlvm.ir.ArrayType)
+                    and len(exponent.type) == 1
+                    and not isinstance(exponent.type.element, pnlvm.ir.ArrayType)
+                ):
+                    exponent = b.extract_value(exponent, [0])
+                else:
+                    # variable (input) matching array specification
+                    exponent = self._gen_llvm_load_param(ctx, b, params, EXPONENTS, in_idx, 1.0)
+
+                # FIXME: Remove this micro-optimization,
+                #        it should be handled by the compiler
+                if not isinstance(exponent, pnlvm.ir.Constant) or exponent.constant != 1.0:
+                    in_val = b.call(pow_f, [in_val, exponent])
+
+                # lower-dim array specification (see exponent above)
+                weight = self._gen_llvm_load_param(ctx, b, params, WEIGHTS, idx, 1.0)
+                if (
+                    isinstance(weight.type, pnlvm.ir.ArrayType)
+                    and len(weight.type) == 1
+                    and not isinstance(weight.type.element, pnlvm.ir.ArrayType)
+                ):
+                    weight = b.extract_value(weight, [0])
+                else:
+                    # variable (input) matching array specification
+                    weight = self._gen_llvm_load_param(ctx, b, params, WEIGHTS, in_idx, 1.0)
+
+                in_val = b.fmul(in_val, weight)
+
+                # val_p points to a temp value that will be put in a specific
+                # element in the result. It's calculated sequentially by
+                # combining each in_val iterated over in this loop
+                val = b.load(val_p)
+                val = getattr(b, comb_op)(val, in_val)
+
+                b.store(val, val_p)
+
+            # value (output) matching
+            scale = self._gen_llvm_load_param(ctx, builder, params, SCALE, vo_idx, 1.0)
+            offset = self._gen_llvm_load_param(ctx, builder, params, OFFSET, vo_idx, -0.0)
+
+            val = builder.load(val_p)
+            val = builder.fmul(val, scale)
+            val = builder.fadd(val, offset)
+
+            builder.store(val, ptro)
+
+    def _gen_llvm_combine(self, builder, ctx, vi, vo, params):
         # assume operation does not change dynamically
         operation = self.parameters.operation.get()
         if operation == SUM:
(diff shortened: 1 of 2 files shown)


psyneulink/core/components/functions/nonstateful/transformfunctions.py
```diff
@@ -1524,10 +1529,79 @@ def _get_input_struct_type(self, ctx):
...
-        offset = self._gen_llvm_load_param(ctx, builder, params, OFFSET, index, -0.0)
+    def _gen_llvm_function_body(self, ctx, builder, params, _, arg_in, arg_out, *, tags: frozenset):
+        # Sometimes we arg_out to 2d array
+        if self.defaults.variable.ndim == self.defaults.value.ndim:
+            arg_out = pnlvm.helpers.unwrap_2d_array(builder, arg_out)
+
+        self._gen_llvm_combine(builder, ctx=ctx, vi=arg_in, vo=arg_out, params=params)
+        return builder
+
+    def _llvm_combine_body(self, builder, ctx, vi, vo, val_f, pow_f, comb_op, params, indices):
```
jvesely: All code generating functions should be prefixed with `_gen_llvm`

psyneulink/core/components/functions/nonstateful/transformfunctions.py
```diff
@@ -103,8 +103,13 @@ def _gen_llvm_load_param(self, ctx, builder, params, param_name, index, default)
             return ctx.float_ty(default)
 
         elif isinstance(param_type, pnlvm.ir.ArrayType):
-            index = ctx.int32_ty(0) if len(param_type) == 1 else index
-            param_ptr = builder.gep(param_ptr, [ctx.int32_ty(0), index])
+            if len(param_type) == 1 and not isinstance(param_type.element, pnlvm.ir.ArrayType):
+                index = [ctx.int32_ty(0)]
+
+            if not isinstance(index, list):
```
jvesely, line 118: It might be better to guarantee that index/indices is always a list.

psyneulink/core/components/functions/nonstateful/transformfunctions.py
```diff
@@ -1524,10 +1529,79 @@ def _get_input_struct_type(self, ctx):
         default_var = np.atleast_2d(self.defaults.variable)
         return ctx.convert_python_struct_to_llvm_ir(default_var)
 
-    def _gen_llvm_combine(self, builder, index, ctx, vi, vo, params):
-        scale = self._gen_llvm_load_param(ctx, builder, params, SCALE, index, 1.0)
-        offset = self._gen_llvm_load_param(ctx, builder, params, OFFSET, index, -0.0)
+    def _gen_llvm_function_body(self, ctx, builder, params, _, arg_in, arg_out, *, tags: frozenset):
+        # Sometimes we arg_out to 2d array
+        if self.defaults.variable.ndim == self.defaults.value.ndim:
```
jvesely: It'd be better to use compiled data types rather than python dimensionality to maintain a single source of truth.

psyneulink/core/components/functions/nonstateful/transformfunctions.py
```diff
@@ -1524,10 +1529,79 @@ def _get_input_struct_type(self, ctx):
...
+
+        self._gen_llvm_combine(builder, ctx=ctx, vi=arg_in, vo=arg_out, params=params)
+        return builder
+
+    def _llvm_combine_body(self, builder, ctx, vi, vo, val_f, pow_f, comb_op, params, indices):
+        ptro = builder.gep(vo, [ctx.int32_ty(0), *indices])
+
+        if isinstance(ptro.type.pointee, pnlvm.ir.ArrayType):
+            with pnlvm.helpers.array_ptr_loop(builder, ptro, f"combine_axis{len(indices)}") as (b, idx):
+                self._llvm_combine_body(b, ctx, vi, vo, val_f, pow_f, comb_op, params, [*indices, idx])
```
jvesely: Rather that recursing here, it might be clearer to use `recursive_iterate_arrays` over the output array.

tests/functions/test_combination.py
```diff
@@ -270,6 +285,54 @@ def test_linear_combination_function(variable, operation, exponents, weights, sc
...
+@pytest.mark.parametrize("weights", [None, 0.5, 'A'], ids=["W_NONE", "W_SCALAR", "W_ARRAY"])
+@pytest.mark.parametrize("scale", [None, RAND1_S, 'A'], ids=["S_NONE", "S_SCALAR", "S_ARRAY"])
+@pytest.mark.parametrize("offset", [None, RAND2_S, 'A'], ids=["O_NONE", "O_SCALAR", "O_ARRAY"])
+def test_linear_combination_function_higher_dim(variable, operation, exponents, weights, scale, offset, func_mode, benchmark):
+    # arrays in shape of input
+    if weights == 'A':
+        # random 1/-1
+        weights = 2 * (np.round(RANDh_V['weights'][variable.shape]) - .5)
+    if exponents == 'A':
+        exponents = RANDh_V['exponents'][variable.shape]
```
jvesely: Why can't the above be passed as parametrized arguments?

kmantel: They could be generated in the test too. 

Otherwise, I think the issue is that they need to match [parts of] the variable shape, but AFAIK there isn't a way to couple parametrization vars (as in, only do `(var1,exp1), (var2,exp2)...` and not also `(var1,exp2), (var2,exp1)`), so I'd still need to pass some function or dict to get the correctly shape value out.

But yeah I can also pre-calculate these a bit better.

psyneulink/core/components/functions/nonstateful/transformfunctions.py
kmantel: Can this be used when `exponents`/`weights` are either scalar or matching the input shape? Not sure about it, but it seems that since `recursive_iterate_arrays` yields `gep` results, it would need to conditionally include or exclude those parameter vals in the `recursive_iterate_arrays` calls, and need to be repeated per combination.

psyneulink/core/components/functions/nonstateful/transformfunctions.py
kmantel: I did that trying to deal with using lists as index in LinearCombination (ex L1568 `exponent = self._gen_llvm_load_param(ctx, b, params, EXPONENTS, in_idx, 1.0)` without changing other TransformFunctions or overriding `_gen_llvm_load_param` and couldn't think of something better. I'm fine changing it if it's better to do one of those or something else I haven't thought of.

psyneulink/core/components/functions/nonstateful/transformfunctions.py
kmantel: @jvesely sorry, would you say this part is just unclear, or incorrect/costly/risky that it shouldn't wait to be reworked? I was a bit stumped on it considering my comment above. The overall support for >2d depends on these/similar changes

jvesely: Hi, sorry for the delay. I'll take another look at the updated changes today or tomorrow.

psyneulink/core/components/functions/nonstateful/transformfunctions.py
```diff
@@ -1529,10 +1534,81 @@ def _get_input_struct_type(self, ctx):
...
+        # for the frequent case where non-LLVM variable is 2D and value is cast
+        # up to 2D from 1D. arg_out is expected to have one fewer dimension than
+        # arg_in in _gen_llvm_combine_body.
+        if (arg_in.type.pointee.element == arg_out.type.pointee.element) and arg_out.type.pointee.count == 1:
+            arg_out = pnlvm.helpers.unwrap_2d_array(builder, arg_out)
+
+        self._gen_llvm_combine(builder, ctx=ctx, vi=arg_in, vo=arg_out, params=params)
+        return builder
+
+    def _gen_llvm_combine_body(self, builder, ctx, vi, vo, val_f, pow_f, comb_op, params, indices):
```
jvesely: What is `val_f` ? if it's not a function, it probably shouldn't have the `_f` suffix.

tests/functions/test_combination.py
```diff
@@ -270,6 +285,54 @@ def test_linear_combination_function(variable, operation, exponents, weights, sc
...
+
+    scale = 1.0 if scale is None else scale
+    offset = 0.0 if offset is None else offset
+    exponent = 1.0 if exponents is None else exponents
+    weights = 1.0 if weights is None else weights
+
+    tmp = (variable ** exponent) * weights
+    if operation == pnl.SUM:
+        expected = np.sum(tmp, axis=0) * scale + offset
+    if operation == pnl.PRODUCT:
```
jvesely: might need to use `elif`, including
```
elif
    assert False, "Unknown operation"
```
at the end.

tests/functions/test_combination.py
```diff
@@ -270,6 +285,54 @@ def test_linear_combination_function(variable, operation, exponents, weights, sc
...
+        expected = np.prod(tmp, axis=0) * scale + offset
+
+    # wider tolerances needed for fp32
+    np.testing.assert_allclose(res, expected, rtol=3e-5, atol=2e-7)
```
jvesely: If the more lenient tolerance is only needed for fp32, please apply it only for fp32.

tests/functions/test_combination.py
```diff
@@ -178,6 +178,21 @@ def test_matrix(self):
 RAND3_S = np.random.rand()
 
 
+# higher dimension arrays
+test_varh1 = np.random.rand(1, 1, SIZE)
+test_varh2 = np.random.rand(2, 3, SIZE, SIZE)
+test_varh3 = np.random.rand(5, 4, SIZE, SIZE, SIZE)
```
jvesely: Does using float32 dtype for the randomly generated values work to avoid having to relax tolerances when checking results?

kmantel: Ah, yeah it does. I do have to do this for all of the random vars though, or some other variants of other tests here will go out of tolerance (at least `test_reduce_function` but I haven't run enough to see if it's always the same)

psyneulink/core/components/functions/nonstateful/transformfunctions.py
kmantel: Changed; it's not a function, I just misunderstood the naming convention I think

tests/functions/test_combination.py
kmantel: Done here, though I copied this chunk from the other tests in test_combination.py and didn't touch those

psyneulink/core/components/functions/nonstateful/transformfunctions.py
```diff
@@ -1529,10 +1534,81 @@ def _get_input_struct_type(self, ctx):
...
+        if isinstance(ptro.type.pointee, pnlvm.ir.ArrayType):
+            with pnlvm.helpers.array_ptr_loop(builder, ptro, f"combine_axis{len(indices)}") as (b, idx):
+                self._gen_llvm_combine_body(b, ctx, vi, vo, val, pow_f, comb_op, params, [*indices, idx])
+        else:
+            val_p = builder.alloca(val.type, name="combined_result")
+            builder.store(val, val_p)
+
+            assert isinstance(vi.type.pointee, pnlvm.ir.ArrayType)
+            with pnlvm.helpers.array_ptr_loop(builder, vi, "combine") as (b, idx):
+                in_idx = [idx, *indices]
```
jvesely: how come "idx" is prepended to the list of indices? I'd expect the innermost loop to change the last index.

tests/functions/test_combination.py
```diff
@@ -152,30 +152,55 @@ def test_matrix(self):
     #
 
 
+def _rand(*args) -> np.ndarray:
+    res = np.random.rand(*args)
+    if pytest.helpers.llvm_current_fp_precision() == 'fp32':
+        try:
+            res = res.astype(np.float32)
+        except AttributeError:
```
jvesely: does `np.random.default_rng().random(size=(*args), dype=np.float32)` work without the extra scalar handling logic?

psyneulink/core/components/functions/nonstateful/transformfunctions.py
kmantel: `indices` points to the output result; I believe doing `[indices, idx]` would be an `axis=-1` combination, but we do axis 0, so `idx` corresponds to the outermost index in the input array.
I added some more comments as well, maybe too verbose because I also had to reclarify to myself why it's this way.

I also renamed `indices` to `vo_idx`, but I can change it back or to something else if you think that's better.

tests/functions/test_combination.py
kmantel: I tried with

```diff
diff --git a/tests/functions/test_combination.py b/tests/functions/test_combination.py
index 1910530154..506c952a1b 100644
--- a/tests/functions/test_combination.py
+++ b/tests/functions/test_combination.py
@@ -153,13 +153,12 @@ class TestReduce:
 
 
 def _rand(*args) -> np.ndarray:
-    res = np.random.rand(*args)
     if pytest.helpers.llvm_current_fp_precision() == 'fp32':
-        try:
-            res = res.astype(np.float32)
-        except AttributeError:
-            res = np.float32(res)
-    return res
+        dtype = np.float32
+    else:
+        dtype = np.float64
+
+    return np.random.default_rng().random(size=args, dtype=dtype)
 
 
 SIZE=5
```

but it seems to go back to failing tolerances for some reason

kmantel: I think the tolerance failures were just intermittent and possibly due to the numbers directly in the parametrizations. I added conversions both here and in the test using `np.asarray` which seems to be consistent.

PR #3547, opened 2026-05-27
AutodiffComposition: Construct pytorch representation in compiled forward path

jvesely:
All structures need to be finalized before compilation starts.

### DIFF
--- a/psyneulink/core/llvm/codegen.py
+++ b/psyneulink/core/llvm/codegen.py
@@ -1239,9 +1239,9 @@ def gen_autodiffcomp_exec(ctx, composition, *, tags:frozenset):
     """Creates llvm bin execute for autodiffcomp"""
     assert composition.controller is None
 
-    context = composition._context_for_pytorch
-    composition._build_pytorch_representation(context=context, skip_backprop_unless_learning=True)
-    pytorch_model = composition.parameters.pytorch_representation.get(context)
+    pytorch_model = composition.parameters.pytorch_representation.get(composition._context_for_pytorch)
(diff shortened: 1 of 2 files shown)


kmantel (approved):
This is good. 

Just for context from what I understood about the problem, calling `_build_pytorch_representation` in `AutodiffComposition.run` makes sure that the new CIM input/output ports corresponding to the generated backprop mechs exist by the time `_initialize_from_context` is called in `Composition.run`, so that their `variable`/`value` values in `context` are not set with the incorrect shape.

jvesely:
> Just for context from what I understood about the problem, calling `_build_pytorch_representation` in `AutodiffComposition.run` makes sure that the new CIM input/output ports corresponding to the generated backprop mechs exist by the time `_initialize_from_context` is called in `Composition.run`, so that their `variable`/`value` values in `context` are not set with the incorrect shape.

I'm not sure about the pnl aspects of calling different setup/initialization routines. I wouldn't expect it to adjust shapes based on backprop pathways. The call to `infer_backpropagation_learning_pathways` is skipped if the forward execution is called before learn, and the backprop adjustments should already be done if we call forward execution after learn.

The intention here was to move `_build_pytorch_representation` out of the compilation paths and make it explicit that compiled forward execution needs pytorch representation.
I tried to model it after the 'learn' block for llvm and pytorch just above, but it needed a few extra steps to pass the tests.
The need for these extra initialization steps in the newly added block (`_analyze_graph`, `_initialize_from_context`) is not quite clear to me.

PR #3560, opened 2026-06-17
Fix FitzHughNagumoIntegrator reset crash and value-shape inconsistency in a Mechanism

davidt0x:
## Summary

Fixes two related bugs affecting `FitzHughNagumoIntegrator` (and multi-stateful integrators generally) when used inside a `Mechanism`. Both surfaced from the reproducer below.

### Reproducer

```python
import psyneulink as pnl

inp = pnl.ProcessingMechanism(input_shapes=1)
lc = pnl.TransferMechanism(
    input_shapes=1,
    integrator_mode=True,
    integrator_function=pnl.FitzHughNagumoIntegrator(),
    function=pnl.Linear(),
    output_ports=[{pnl.VARIABLE: (pnl.OWNER_VALUE, 1)}],
    reset_stateful_function_when=pnl.AtTrialStart(),
)
comp = pnl.Composition()
comp.add_nodes([inp, lc])
comp.add_projection(sender=inp, receiver=lc)
comp.run(inputs={inp: [[1.0]]})   # IndexError in Python mode; worked in LLVMRun
```

## Bug 1 — `IndexError` on reset (`mechanism.py`)

On reset, the Mechanism rebuilt its value from the integrator's reset output but kept only element `[0]`:

```python
new_input = self.integrator_function.reset(...)[0]
```

`StatefulFunction.reset()` returns one entry per stateful attribute. For single-attribute integrators (`AdaptiveIntegrator` → `[previous_value]`) the `[0]` correctly unwraps. For `FitzHughNagumoIntegrator` → `[v, w, time]` it discarded `w` and `time`, leaving the value with a single element, so an `OutputPort` selecting `(OWNER_VALUE, 1)` indexed past the end → `IndexError`. Compiled mode was unaffected (statically-typed value buffer; reset only rewrites the integrator state).

**Fix:** keep all stateful terms when there is more than one.

## Bug 2 — value-shape inconsistency between Python and compiled (`integratorfunctions.py`)

Even after Bug 1, the mechanism's value shape differed between modes: Python produced `(3, 1, 1)`, compiled `(3, 1)` (and `defaults.value` was `(3, 1)`). FHN stacks its three stateful terms; in a Mechanism, `variable` is the Mechanism's 2d variable `(1, 1)` while the stateful values are 1d, so broadcasting during execution inflated each term to 2d → `(3, 1, 1)`. During initialization (which sizes `value`) the un-broadcast 1d state is returned → `(3, 1)`. Compiled unwraps the same leading dimension (`helpers.unwrap_2d_array`), so it stays `(3, 1)`.

**Fix:** conform `variable` to the rank of the stateful values at the start of `_function` (drop the leading singleton input-port dimension), so Python matches both the advertised default and compiled execution at `(3, 1)`. Standalone use with a genuinely 2d variable (and 2d stateful values) is unaffected. This required no compiled-codegen changes.

## Testing

- `test_reset_multi_stateful_integrator_with_owner_value_output_port` runs across Python and all compiled modes, asserts the value shape is `(3, 1)`, and asserts a shared value reference (tolerance covers fp32 compiled CI).
- `test_FitzHughNagumo_simple_scalar` previously added a dimension to the Python expected result to work around the shape mismatch; that workaround is removed (Python now matches compiled).
- `test_FitzHughNagumo_valid` and the gilzenrat figure-2 test were updated from the inflated `(3, 1, 1)` indexing to `(3, 1)` (values unchanged).
- Full FHN / transfer / integrator / processing / LCA / mdf suites pass in Python and compiled.

## Notes

- `OrnsteinUhlenbeckIntegrator` has a similar shape inconsistency (plus extra quirks and no existing compiled coverage). It is intentionally left out of this PR and tracked separately.
- The construction-dependent FHN shape (`(3,1,1)` when the class, rather than an instance, is passed) is a separate pre-existing issue (the existing `xfail`) and is untouched here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/mechanisms/test_transfer_mechanism.py
```diff
@@ -1830,6 +1830,39 @@ def test_integrator_mode_reset_in_composition(comp_mode):
...
+    comp = pnl.Composition()
+    comp.add_nodes([inp, lc])
+    comp.add_projection(sender=inp, receiver=lc)
+
+    comp.run(inputs={inp: [[1.0]]}, execution_mode=comp_mode)
+
+    # The mechanism's value is the 3 FHN terms (v, w, time), each of size 1, so
+    # its shape is (3, 1) in both Python and compiled mode (previously Python
+    # inflated this to (3, 1, 1); see FitzHughNagumoIntegrator._function).
+    np.testing.assert_array_equal(np.asarray(lc.parameters.value.get(comp)).shape, (3, 1))
```
kmantel: I think if `lc.parameters.value.get(comp)` is not already a numpy array, this is probably a bug.

psyneulink/core/components/mechanisms/mechanism.py
```diff
@@ -2354,7 +2354,19 @@ def reset(self, *args, force=False, context=None, **kwargs):
...
+                # with a single stateful_attribute (the common case), unwrap the
+                # list so the function operates on that value directly.  For
+                # integrators with multiple stateful_attributes, pass all of them
+                # so the resulting value matches the multi-element value produced
+                # during normal execution (see _parse_function_variable).
+                reset_value = self.integrator_function.reset(*args, **kwargs, context=context)
+                if len(reset_value) == 1:
+                    new_input = reset_value[0]
+                else:
+                    new_input = convert_all_elements_to_np_array(reset_value)
```
kmantel, line 2369: I would expect that `reset_value` should already be (and contain if ragged) a numpy array at this point. Are there cases where it's not? If so, I think that's probably a bug and should probably be handled in `reset`.

psyneulink/core/components/functions/stateful/integratorfunctions.py
```diff
@@ -4886,6 +4886,20 @@ def _function(self,
...
+        # integrator_function of a Mechanism, variable is the Mechanism's 2d
+        # variable (num_input_ports x port_size) and so carries an extra leading
+        # (input-port) dimension relative to the 1d stateful values (previous_v,
+        # etc.). Broadcasting the 1d stateful values against that 2d variable
+        # would otherwise inflate the result with an extra dimension (e.g.
+        # (3, 1, 1) instead of (3, 1)), inconsistent with the value computed
+        # during initialization and with compiled execution (which unwraps the
+        # same dimension; see helpers.unwrap_2d_array). Standalone use with a
+        # genuinely 2d variable (matching 2d stateful values) is unaffected.
+        if not np.isscalar(variable) and variable.ndim > np.ndim(previous_v) \
```
kmantel: Shouldn't `variable.ndim > np.ndim(previous_v)` already imply that `not np.isscalar(variable)`?

psyneulink/core/components/functions/stateful/integratorfunctions.py
```diff
@@ -4886,6 +4886,20 @@ def _function(self,
...
+                and variable.size == np.size(previous_v):
```
kmantel: I think if the sizes are not the same, this is probably a problem and should error. Even if they are the same, I think if their squeezed shapes are different, this *may* be a problem that we want to error, but I could understand if we want to ignore that and proceed anyway.

tests/mechanisms/test_transfer_mechanism.py
```diff
@@ -1830,6 +1830,39 @@ def test_integrator_mode_reset_in_composition(comp_mode):
...
+
+    # Output port selects the second FHN term (w); identical in Python and compiled
+    # (tolerance allows for reduced fp32 precision in compiled CI runs).
+    np.testing.assert_allclose(np.squeeze(comp.results), 0.002795524774843733, rtol=1e-5, atol=1e-8)
```
kmantel: I think it would be better to check the result explicitly without squeeze

kmantel (commented):
As a general comment that's probably be beyond the scope of this bug, I think the `Mechanism` reset chain should probably make of use `_parse_integrator_function_variable` and `_parse_function_variable` (these are not currently implemented for Mechanism). This would formalize the modification in `reset` and also ensure that `FitzHughNagumoIntegrator.function` would be passed input that already has the correct shape during Mechanism execution.

tests/mechanisms/test_integrator_mechanism.py
```diff
@@ -45,29 +45,29 @@ def test_FitzHughNagumo_valid(self):
         I.reset_stateful_function_when = Never()
         I.execute(1.0)
 
-        np.testing.assert_allclose([[0.05127053]], I.value[0], rtol=1e-5, atol=1e-8)
-        np.testing.assert_allclose([[0.00279552]], I.value[1], rtol=1e-5, atol=1e-8)
-        np.testing.assert_allclose([[0.05]], I.value[2], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.05127053], I.value[0], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.00279552], I.value[1], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.05], I.value[2], rtol=1e-5, atol=1e-8)
```
jvesely: It looks like the expected `I.value` is `[[0.05127053], [0.00279552], [0.05]]`. Can this be combined into one line?

tests/mechanisms/test_integrator_mechanism.py
```diff
@@ -45,29 +45,29 @@ def test_FitzHughNagumo_valid(self):
...
         np.testing.assert_allclose(0.01, I.function.value[0], rtol=1e-5, atol=1e-8)
         np.testing.assert_allclose(0.02, I.function.value[1], rtol=1e-5, atol=1e-8)
         np.testing.assert_allclose(0.03, I.function.value[2], rtol=1e-5, atol=1e-8)
 
-        np.testing.assert_allclose([[0.05127053]], I.value[0], rtol=1e-5, atol=1e-8)
-        np.testing.assert_allclose([[0.00279552]], I.value[1], rtol=1e-5, atol=1e-8)
-        np.testing.assert_allclose([[0.05]], I.value[2], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.05127053], I.value[0], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.00279552], I.value[1], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.05], I.value[2], rtol=1e-5, atol=1e-8)
```
jvesely: Same here. Can this be combined into one line?

tests/mechanisms/test_integrator_mechanism.py
```diff
@@ -45,29 +45,29 @@ def test_FitzHughNagumo_valid(self):
...
+        np.testing.assert_allclose([0.05127053], I.output_ports[0].value, rtol=1e-5, atol=1e-8)
 
         I.execute(1.0)
 
-        np.testing.assert_allclose([[0.06075727]], I.value[0], rtol=1e-5, atol=1e-8)
-        np.testing.assert_allclose([[0.02277156]], I.value[1], rtol=1e-5, atol=1e-8)
-        np.testing.assert_allclose([[0.08]], I.value[2], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.06075727], I.value[0], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.02277156], I.value[1], rtol=1e-5, atol=1e-8)
+        np.testing.assert_allclose([0.08], I.value[2], rtol=1e-5, atol=1e-8)
```
jvesely: Same here. Can this be combined into one line?

tests/mechanisms/test_transfer_mechanism.py
jvesely: +1 to Katherine's comment. It should also make explicit testing of shape, above, unnecessary.

psyneulink/core/components/functions/stateful/integratorfunctions.py
davidt0x: Dropped the np.isscalar guard — using np.ndim(variable) handles scalars (ndim 0), so the comparison can't reach a scalar .ndim (#3).

psyneulink/core/components/functions/stateful/integratorfunctions.py
davidt0x: Dropped the size == guard — np.reshape now raises on a genuine size mismatch instead of silently skipping (#4).

tests/mechanisms/test_integrator_mechanism.py
davidt0x: Good catch, fixed. I think AI copied other places in the file that did this. I will leave them for now.

tests/mechanisms/test_integrator_mechanism.py
davidt0x: Fixed

tests/mechanisms/test_integrator_mechanism.py
davidt0x: Fixed

tests/mechanisms/test_transfer_mechanism.py
davidt0x: Good catch — that np.asarray was redundant. I've removed the whole shape-assertion line in favor of an explicit assert_allclose(comp.results, ...) that checks shape and value together (per the comments below). I also confirmed parameters.value.get() returns an ndarray here, so it wasn't masking a bug.

tests/mechanisms/test_transfer_mechanism.py
davidt0x: Fixed

psyneulink/core/components/mechanisms/mechanism.py
davidt0x: So `reset_value` was not already an array: `StatefulFunction.reset` was building and returning a plain Python `list` of the stateful values, so every caller (`Mechanism_Base.reset, Component.reset`, `DDM.reset`) was separately converting it. I moved `convert_all_elements_to_np_array` into `StatefulFunction.reset` itself, so it now returns (and stores in its value `Parameter`) a single array — an object array if the stateful attributes are ragged — and dropped the now-redundant conversions here and in `DDM.reset`. (`Mechanism_Base.reset's convert_to_np_array(..., dimension=2)` and `Component.reset`'s `np.atleast_2d` are kept since they enforce a 2d shape, not just type.) What do you think @kmantel, does this sound cleaner?

tests/mechanisms/test_transfer_mechanism.py
davidt0x: Fixed

psyneulink/core/components/mechanisms/mechanism.py
kmantel: I think the original workaround here was appropriate for now then, because these changes to reset are not really related to this PR's bug.

Just converting the value this way in StatefulFunction.reset adds another conversion step to those in the external sets which would be redoing the conversion multiple times already. Which is undesired if not a bug, but I don't think doing that fits here.

psyneulink/core/components/functions/stateful/integratorfunctions.py
kmantel: Is it an intentional choice to use reshape vs removing a dimension? Just trying to understand.

psyneulink/core/components/mechanisms/mechanism.py
davidt0x: Ok I reverted to the original workaround.

psyneulink/core/components/functions/stateful/integratorfunctions.py
davidt0x: Yes, it's intentional. The goal there is to conform variable to the shape of the integrator's stateful values (previous_v, etc.) — not just to drop a specific axis. In a Mechanism, variable arrives as the Mechanism's 2d variable (num_input_ports × port_size, e.g. (1, 1)) while the stateful values are 1d (e.g. (1,)), so I reshape variable to match np.shape(previous_v).

I used np.reshape(variable, np.shape(previous_v)) rather than removing a dimension (e.g. variable[0] / np.squeeze) for two reasons:

1. It targets the exact expected shape instead of assuming the extra dimension is a leading singleton. Stripping "a" axis only works for that one case; reshaping to the stateful value's shape is unambiguous and also handles e.g. (1, N) → (N,).
2. Mainly though it prevents hiding errors like you asked. It validates rather than hides mismatches. If variable and the stateful values have incompatible element counts, reshape raises — which surfaces a real bug — whereas squeeze/indexing could silently yield a wrong-but-valid shape. (This was in response to an earlier comment about preferring to error on a size mismatch.)

Longer term this fix I guess could be incorporated into `_parse_integrator_function_variable`.

PR #3561, opened 2026-06-18
Distributed (Dask) fitting for ParameterEstimationComposition

aryan-pan:
## Summary

Adds distributed evaluation to `ParameterEstimationComposition` fitting. Candidate
parameterizations are evaluated in parallel across a Dask cluster instead of
serially, controlled by a single constructor flag. Serial behavior is unchanged
when the flag is off, and Dask is imported only when distributed fitting is
requested.

## User-facing API

`PECOptimizationFunction` and `ParameterEstimationComposition` gain two arguments:

- `distributed: bool = False` — evaluate candidates across a Dask cluster.
- `distributed_options: Mapping` — configuration with the following keys:
  - `pec_factory` — required; a picklable callable that takes the observed data and
    returns a fresh `(pec, inputs)` for a worker to score.
  - `worker_cores` — LLVM threads per worker; defaults to `$SLURM_CPUS_PER_TASK`.
  - `max_concurrent_evaluations` — candidates per ask/tell round; defaults to the
    live worker count.
  - `client` — a Dask client to use instead of an auto-resolved cluster.
  - `n_workers` — size of the auto-created single-node `LocalCluster`.

Both the optuna samplers and `differential_evolution` are supported. LLVM
execution and a single scalar objective are required.

## Launch models

- Single node: `python study.py` forms a `LocalCluster` automatically.
- Multiple nodes: `srun -n N python -m psyneulink.dask_run study.py`. A new
  launcher module forms the scheduler, driver, and workers from the SLURM ranks,
  so the study script contains no scheduler addresses or cluster setup. The worker
  count is `N - 2`.

A worked DDM example and an SBATCH template are under `Scripts/Debug/pec_dask/`.

## How it works

The driver owns the optimizer. Each ask/tell round dispatches a batch of
candidates to workers; a worker rebuilds one PEC locally from `pec_factory`,
caches it, and returns `float(pec.log_likelihood(...))`. A live PEC is never
shipped across the wire, since Compositions are not safely picklable. With common
random numbers every candidate is scored against identical simulation noise, so a
distributed fit with a tell-order-independent sampler matches a serial fit.
Without common random numbers the fit is still valid but not reproducible, and a
warning is emitted.

## Included prerequisite

This branch bundles the commit "Fix PEC log_likelihood LLVM evaluation", which
makes `log_likelihood` return a scalar correctly and enforces LLVM mode. The
distributed workers depend on it. A follow-up commit removes the per-call
`pnllvm.cleanup()` that fix introduced in `log_likelihood`, because
`log_likelihood` re-scores one fixed PEC many times and must reuse the compiled
binary; cleaning up per call forces a full recompile every evaluation. The
once-per-`run()` cleanup is left unchanged.

## Validation

- New unit, driver-logic, and guard tests, plus process-based `LocalCluster`
  integration tests: per-candidate log-likelihood parity with serial to
  `rtol=1e-10`, an end-to-end fit with per-worker PEC caching, serial-versus-
  distributed parity under common random numbers, distributed
  `differential_evolution`, and the no-CRN warning. The new test module is 22
  tests.
- The existing PEC test suite passes unchanged.
- Multi-node runs on a SLURM cluster: 8 workers across 2 nodes via the launcher,
  about a 6.8x speedup over a serial fit on the same budget, with threshold and
  non-decision-time recovered to within a few percent of ground truth.

## Packaging

Adds a `dask` extra. Install with `pip install "psyneulink[dask]"`, which pulls in
`dask`, `distributed`, and `dask-jobqueue`. With `distributed=False` none of these
are imported.

## Contents

- Core: `fitfunctions.py`, `parameterestimationcomposition.py`, the new
  `psyneulink/dask_run.py` launcher, `setup.py`, and `dask_requirements.txt`.
- Tests: `tests/composition/pec/test_pec_dask.py` and additions to the PEC test
  module.
- Supporting material under `Scripts/Debug/pec_dask/`: the example and template, a
  model-agnostic benchmark harness, the original prototype, and benchmark figures.

## Limitations

LLVM execution only; a single scalar objective; `pec_factory` is required for
distributed runs; stateful samplers such as CMA-ES and QMC explore different
points under batched ask/tell and so are not bit-identical to the serial driver.


davidt0x (approved):
LGTM, I pushed some cleanup to the PEC docs in general and added a link to the distributed stuff. Nice work!

PR #3571, opened 2026-06-24
Fix PEC degenerate-likelihood warning re-printing stale parameter sets

davidt0x:
# Summary
The Parameter Estimation Composition (PEC) prints a warning summarizing parameter sets that produced degenerate objective-function values (e.g. BadLikelihoodWarning when a simulation has zero variance and the KDE can't be computed). A user reported that this warning repeatedly lists parameter sets that were not just evaluated — pinning the first bad set seen to the top of the log and re-printing it on every subsequent iteration, cluttering the output.

# Root cause
The optimizer callback (progress_callback in both _fit_differential_evolution and _fit_optuna) summarized warnings from a warns_with_params list that was never cleared. Each callback re-printed the entire accumulated history of degenerate parameter sets, so the earliest bad set kept reappearing regardless of what was actually evaluated in the current iteration/trial.

A related latent bug: the per-evaluation warns buffer was only cleared inside the display_iter branch, so with display_iter=False a stale warns[-1] could be mis-attributed to a later (possibly warning-free) parameter set.

# Changes
Extracted the duplicated "print summary + clear" logic into a new _report_degenerate_warnings helper, which clears warns_with_params after printing so each callback only reports sets that went degenerate since the last callback.
Always clear the per-evaluation warns buffer at the end of each evaluation (not just in the display_iter branch), and hoisted the warning check into a single got_pec_warning flag so the captured warning and parameters always come from the same evaluation.
Fixed a missing space in the footer text (fitting todata → fitting to data).

### DIFF
--- a/tests/composition/pec/test_parameterestimationcomposition.py
+++ b/tests/composition/pec/test_parameterestimationcomposition.py
@@ -11,6 +11,7 @@ import psyneulink as pnl
 
 from psyneulink.core.components.functions.nonstateful.fitfunctions import (
     PECOptimizationFunction,
+    BadLikelihoodWarning,
 )
 
 # Set the number of threads to 1 for all tests in this module, tests are often run in parallel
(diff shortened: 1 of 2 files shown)


tests/composition/pec/test_parameterestimationcomposition.py
```diff
@@ -721,3 +722,66 @@ def objective(sim_data):
     assert "sim_data" in captured
     # sim_data shape: (num_trials, num_estimates, num_outcomes)
     assert np.mean(np.std(captured["sim_data"], axis=1)) > 0.0
+
+
+class _RecordingConsole:
```
jvesely, line 727: Why can't this use existing pytest output capture? or the warning contexts?

davidt0x: Yeah, it probably could have but I missed it an merged it. Let me see if I can clean it up in another PR.

PR #3587, opened 2026-07-08
Replace pydocstyle with ruff

jvesely:
pytest-pydocstyle is archived and no longer maintained.


jvesely:
@davidt0x, this is a small subpart of #3363. More checks can be enabled as we fix the code.

PR #3610, opened 2026-07-31
Fix intermittent LLVM JIT crash from missing target data layout on generated modules

davidt0x:
I iterated with Claude to reproduce, diagnose, and fix the intermittently crashing `test_greedy_agent.py::test_predator_prey` test that has been failing lately. Not exactly sure why it started to manifest recently (maybe the recent llvmlite version bump has perturbed codegen) but I think Claude might have found the fix. Getting it to reproduce was a bit of a pain. Thought it was windows only and kept failing to get a core dump on windows in CI. Then realized it happens on linux as well but it is a bit rarer. Had Claude instrument a modified CI workflow, run a bunch of parallel ubuntu jobs that would dump the core and upload it on a crash. Once it had a core it seemed to diagnose the issue pretty quickly. Seems like a reasonable diagnosis to me but I defer to your expertise @jvesely to make sure I am not missing something important. If you want to play around with the branch I used to reproduce things (`diag/windows-predator-prey-crash`) should reproduce it pretty reliably. The branch is poorly named now though since it actually runs only linux jobs now. 

## Summary

Fixes the intermittent crash in compiled execution — most visibly `tests/models/test_greedy_agent.py::test_predator_prey[..._LLVMPerNode]` — which failed with `Windows fatal exception: access violation` on Windows and `SIGSEGV` on Linux, in `_execute_controller → execute_node → compiled __call__`. It reproduced across platforms and fp precisions, was data/allocator-dependent (only under a churned stack, e.g. deep in a full CI worker), and never showed up in isolated runs.

## Root cause

Generated IR modules were compiled with an **empty data layout** and an `unknown-unknown-unknown` triple (the target was never set on them — this dates back to 2019). With AVX enabled from the host CPU features, the optimizer/vectorizer therefore ran **without the target's stack-alignment spec** (`S128` = 16-byte stack) and emitted **32-byte-aligned** AVX spills to the stack:

```asm
vmovapd %ymm0, -0x80(%rbp)     ; requires 32-byte alignment
```

in a frame whose prologue only maintained the ABI-standard **16-byte** alignment (no `and $-32,%rsp` realignment). The SysV/Windows ABIs guarantee only 16-byte stack alignment, so after `push %rbp` the slot is 32-byte aligned only when the caller's `rsp` happens to land the right way. When a compiled node was invoked (via ctypes) with a 16-but-not-32-byte-aligned stack, the aligned store raised `#GP` → SIGSEGV / access violation.

This explains every observed symptom:
- **Intermittent** — depends on `rsp mod 32` at the ctypes→JIT call, which shifts with accumulated stack state.
- **Platform- and precision-independent** — same AVX codegen on Windows/Linux, fp32/fp64.
- **Invisible to Valgrind** — memcheck doesn't fault misaligned *aligned*-moves.
- **Absent at `opt=0`** — no vectorizer → no aligned `ymm` stack spills.

## Fix

The generated `ir.Module` is **target-neutral by design** — it is built once and compiled by whichever engine is active: the CPU (MCJIT → x86) engine, and when a GPU is present the PTX (→ nvptx64) engine, each of which re-parses the module. The target is a property of the *compilation*, not of the module.

So set `module.triple` and `module.data_layout` **per engine, at optimize time** — in `jit_engine.opt_and_add_bin_module`, the single point every engine passes through, right before the optimization passes run — sourced from each engine's own target machine (CPU → process triple + x86 data layout; PTX → `nvptx64-nvidia-cuda` + NVPTX data layout). With the correct layout the backend emits **unaligned** `ymm` moves (`vmovupd`), which are safe at any alignment.

Applying this per engine (rather than stamping a CPU target onto the shared IR module) keeps the CPU and PTX targets from clobbering each other's layout, and — because it lives on the compile path — no longer constructs a `TargetMachine` on every `LLVMBuilderContext` creation (including pure-Python use that never compiles).

**Runtime type/struct layout is unchanged** — it is the same layout the backend already used at codegen; we only give the optimizer the target info it was missing.

## Validation

- Generated asm for the crashing node: aligned `ymm`→stack stores drop **266 → 0** (replaced by `vmovupd`).
- 179 compiled-mode CPU tests pass with identical numerical results; struct size-match assertions hold.
- End-to-end: a reproduction harness that reliably produced a core on every push produced **zero crashes across all legs** with this change.
- **CUDA/PTX validated on GPU** (RTX 2080 Ti): the PTX path emits byte-identical code with the corrected per-engine target, and 1643 PTX tests pass (0 failures) — confirming the per-engine change is a no-op for the GPU path (no regression).

## Also — `tests/loss` fp32 tolerance

Giving the optimizer the target data layout legitimately shifts some fp32 results by a ULP (correct codegen vectorizes slightly differently). `test_loss` compared compiled results at the default `rtol=1e-7` (≈ float32 epsilon) — the L0 case even hard-codes `expected` to the midpoint of the specific fp32/fp64 results — leaving essentially zero margin, so a single-ULP change tips it over. Relaxed the tolerance to `rtol=1e-5, atol=1e-8` for **compiled fp32 runs only**; Python and fp64 keep the strict default.

## Notes

- Bisects to the recent llvmlite/LLVM bump (`<0.48 → <0.49`, #3592): the empty-data-layout bug is old and latent; newer LLVM's codegen is what began emitting the aligned `ymm` stack spills that expose it.

<details>
<summary>Crash forensics (extracted from a Linux core dump)</summary>

From a core of the crash (Python 3.12, `--fp-precision=fp32`, ubuntu CI runner).

> The raw core (2.2 GB) is intentionally not attached: it is a full process memory image and would contain CI environment secrets (e.g. `GITHUB_TOKEN`), and it is only interpretable with that runner's exact binaries/`.so`s. The small, portable evidence below is extracted from it.

**Native backtrace (faulting thread)**

```
Program terminated with signal SIGSEGV, Segmentation fault.

Thread 1 (LWP 3443):
#0  pthread_kill ()                 from libc.so.6
#1  raise ()                        from libc.so.6
#2  <signal handler called>              # faulthandler re-raising the SIGSEGV
#3  0x00007f85f002cb5a in ?? ()          # <-- FAULT SITE: JIT code (anon exec mapping, no symbol)
#4  0x3ff285bfe2d97e8c in ?? ()          # 0x3ff…/0x3fe…/0xbfe… = IEEE-754 doubles left on the
#5  0x3fe3734561deed36 in ?? ()          #   stack; the JIT frame has no CFI, so gdb's naive scan
...                                      #   walks into numeric data, not real return addresses
#17 llvm::DomTreeBuilder::SemiNCAInfo<…MachineBasicBlock…>::getChildren   # stale codegen frame (red herring)
```

The Python-level fault handler independently reported the same site: `_execute_controller → execute_node → llvm/__init__.py:200 (compiled __call__)` — i.e. **execution of a compiled node**, not compilation.

**Faulting instruction (disassembled from the core at PC `0x7f85f002cb5a`)**

```asm
0x7f85f002cb36:  vbroadcastsd 0x158(%r12),%ymm2      ; vectorized FP, assembling a ymm
0x7f85f002cb40:  vblendpd $0x8,%ymm2,%ymm1,%ymm1     ;   from param/data struct fields (%r12)
0x7f85f002cb46:  vmovddup 0x178(%r12),%xmm2
0x7f85f002cb50:  vblendpd $0x1,%ymm2,%ymm1,%ymm1
0x7f85f002cb56:  vaddpd %ymm0,%ymm1,%ymm0
0x7f85f002cb5a:  vmovapd %ymm0,-0x80(%rbp)           ; <-- FAULT: 32-byte ALIGNED store to a stack slot
```

`vmovapd` (aligned 256-bit AVX store) requires 32-byte alignment or it raises `#GP` → `SIGSEGV`. The target is a stack slot (`-0x80(%rbp)`).

**The enclosing function never realigns the stack**

Prologue of the function containing the fault (`0x7f85f002bd70`):

```asm
0x7f85f002bd70:  push %rbp
0x7f85f002bd71:  mov  %rsp,%rbp
0x7f85f002bd74:  push %r15 / %r14 / %r13 / %r12 / %rbx
0x7f85f002bd7d:  sub  $0xd8,%rsp                     ; NO `and $-32,%rsp`  -> frame is only 16B-aligned
```

Other functions in the *same* JIT module do realign (`and $0xffffffffffffffe0,%rsp`), so the codegen is inconsistent. After `push %rbp`, `rbp mod 32 ∈ {0,16}` depending on the caller's `rsp`; when it is 16, `-0x80(%rbp)` is 16-but-not-32-aligned and the aligned `vmovapd` faults.

**Fix effect on generated code (same test/param)**

| generated asm                              | before fix | after fix |
|--------------------------------------------|-----------:|----------:|
| aligned `vmovapd %ymm` → stack             | 266        | **0**     |
| aligned `vmovapd %ymm` → any memory        | 312        | **0**     |
| unaligned `vmovupd %ymm`                    | 894        | 1239      |

No aligned 32-byte stack store remains, so the fault cannot occur.

</details>


psyneulink/core/llvm/jit_engine.py
```diff
@@ -254,6 +254,17 @@ def __del__(self):
...
     def opt_and_add_bin_module(self, module):
+        # Tag the module with this engine's target triple and data layout before
+        # optimizing. Generated modules carry no data layout of their own, so
+        # without this the optimizer/vectorizer runs with an empty data layout
+        # and can miscompile stack alignment of vectorized (AVX) code -- emitting
+        # a 32-byte-aligned ymm spill (vmovapd) into a stack frame that is only
+        # realigned to the ABI's 16 bytes, an intermittent SIGSEGV/access
+        # violation. Doing this per engine (rather than on the shared IR module)
+        # keeps the CPU and PTX targets from clobbering each other's layout.
+        module.triple = self._target_triple
```
jvesely: This should be able to reuse `self._target_machine.triple`

psyneulink/core/llvm/jit_engine.py
```diff
@@ -369,6 +380,9 @@ def compile_staged(self):
 
 class cpu_jit_engine(jit_engine):
 
+    # Triple of the current process, matching _cpu_jit_constructor's target machine.
+    _target_triple = binding.get_process_triple()
```
jvesely: Then this can be dropped.

psyneulink/core/llvm/jit_engine.py
```diff
@@ -396,6 +410,10 @@ def _init(self):
 
 
 class ptx_jit_engine(jit_engine):
+
+    # Matches _ptx_jit_constructor's target machine triple.
+    _target_triple = "nvptx64-nvidia-cuda"
```
jvesely: and this can be dropped

tests/functions/test_loss.py
```diff
@@ -18,6 +18,16 @@
 
 GROUP_PREFIX="SelectionFunction "
 
+
+def _tolerance_kwargs(mode):
+    # Compiled fp32 execution accumulates rounding error well beyond the default
+    # rtol=1e-7 (which is ~= float32 epsilon), so a single-ULP difference -- e.g.
+    # from a codegen change -- is enough to fail the default comparison. Relax
+    # the tolerance for compiled fp32 runs; Python and fp64 keep the default.
+    if mode != 'Python' and pytest.helpers.llvm_current_fp_precision() == 'fp32':
```
jvesely: `not mode.is_compiled()` is more robust wrt other modes, even if they shouldn't appear in this test.

psyneulink/core/llvm/jit_engine.py
```diff
@@ -254,6 +254,17 @@ def __del__(self):
...
 
     def opt_and_add_bin_module(self, module):
+        # Tag the module with this engine's target triple and data layout before
+        # optimizing. Generated modules carry no data layout of their own, so
+        # without this the optimizer/vectorizer runs with an empty data layout
+        # and can miscompile stack alignment of vectorized (AVX) code -- emitting
+        # a 32-byte-aligned ymm spill (vmovapd) into a stack frame that is only
+        # realigned to the ABI's 16 bytes, an intermittent SIGSEGV/access
+        # violation. Doing this per engine (rather than on the shared IR module)
+        # keeps the CPU and PTX targets from clobbering each other's layout.
```
jvesely: The comment is not useful and should be removed.
Setting the triple and datalayout does a lot more than the single issue that motivated this change.

jvesely (changes requested):
I had a similar change staged for some time: https://github.com/jvesely/PsyNeuLink/commit/e31880d7b79328a0c87cc0bb25361083d2a9202b

The reason that it's not merged is that it introduces quite severe perf degradation. the `ymm→stack stores drop 266 → 0` explains at least part of it. but I think it disables vector instructions in general.

It might be enough to just make sure the invoked function make the right assumption about stack alignment.
 cfunc uses CDECL convention, so we need an invocation wrapper that explicitly sets that (has been on todo list for some https://github.com/orgs/PrincetonUniversity/projects/33/views/1?pane=issue&itemId=73743728)

psyneulink/core/llvm/jit_engine.py
davidt0x: Cleaned up

tests/functions/test_loss.py
davidt0x: mode.is_compiled() doesn't work in this test: func_mode/mech_mode are the string fixtures ('Python'/'LLVM'/'PTX'), not ExecutionMode enums (that's why is_compiled() is fine in test_greedy_agent, where the fixture is a real ExecutionMode). So I went with mode in ('LLVM', 'PTX')

psyneulink/core/llvm/jit_engine.py
davidt0x: Done

psyneulink/core/llvm/jit_engine.py
davidt0x: Good point.

psyneulink/core/llvm/jit_engine.py
davidt0x: Done

davidt0x:
> I had a similar change staged for some time: [jvesely@e31880d](https://github.com/jvesely/PsyNeuLink/commit/e31880d7b79328a0c87cc0bb25361083d2a9202b)
> 
> The reason that it's not merged is that it introduces quite severe perf degradation. the `ymm→stack stores drop 266 → 0` explains at least part of it. but I think it disables vector instructions in general.
> 
> It might be enough to just make sure the invoked function make the right assumption about stack alignment. cfunc uses CDECL convention, so we need an invocation wrapper that explicitly sets that (has been on todo list for some https://github.com/orgs/PrincetonUniversity/projects/33/views/1?pane=issue&itemId=73743728)

Taking a look at your patch it seems to only set the triple. This one sets the data layout as well. Just setting the triple seems to lead to the crash still, gave it a try [here](https://github.com/PrincetonUniversity/PsyNeuLink/actions/runs/30652239773/job/91228052532.). 

Also, I am not seeing the performance drop on my end but maybe I am benchmarking the wrong thing? Tried stability flexibility and predator prey. On my end (just one machine Windows running WSL) it isn't removing vectorization entirely. In fact, looking at the assembly, unaligned vmovupd count actually goes up (894→1239), it's just aligned→unaligned, which is apparently free on modern x86. 

I think your CDECL/alignstack invocation-wrapper idea is still the cleaner long-term design though. However, unless you can confirm the performance degradation maybe we can merge this in the meantime? Maybe the performance regression is dependent on and older version of llvmlite? I didn't test that.

jvesely:
> Taking a look at your patch it seems to only set the triple. This one sets the data layout as well. Just setting the triple seems to lead to the crash still, gave it a try [here](https://github.com/PrincetonUniversity/PsyNeuLink/actions/runs/30652239773/job/91228052532.).
> 
> Also, I am not seeing the performance drop on my end but maybe I am benchmarking the wrong thing? Tried stability flexibility and predator prey. On my end (just one machine Windows running WSL) it isn't removing vectorization entirely. In fact, looking at the assembly, unaligned vmovupd count actually goes up (894→1239), it's just aligned→unaligned, which is apparently free on modern x86.

Ah, thanks for checking the performance. The degradation was observed when I wrote that change, so back in Dec '25 with still using the older llvmlite. 

> I think your CDECL/alignstack invocation-wrapper idea is still the cleaner long-term design though. However, unless you can confirm the performance degradation maybe we can merge this in the meantime?

Agreed. Even if there was perf impact, I'd say that correctness/not crashing is higher priority.
There are other backend things on the todo list like switching to orcjit instead of mcjit, that can all be done later.

The change looks good. 

> Maybe the performance regression is dependent on and older version of llvmlite? I didn't test that.

Agreed. It looks like the new llvmlite using LLVM 22 both changed how vectorization works and both caused this issue and removed the observed perf degradation. Back at the time it was ~2x for stability flexibility model. I don't think we need to care much about perf with old llvmlite. If someone wants better perf they should update to the latest stack.

psyneulink/core/llvm/jit_engine.py
```diff
@@ -396,6 +402,7 @@ def _init(self):
 
 
 class ptx_jit_engine(jit_engine):
+
```
jvesely, line 405: nit: spurious whitespace change

jvesely (approved):

PR #3614, opened 2026-08-04
Require Python >= 3.11

davidt0x:
Python 3.8 (EOL Oct 2024), 3.9 (EOL Oct 2025), and 3.10 are no longer supported. Testing them meant testing a substantially different program: resolving requirements.txt on 3.8 pinned numpy 1.24.4, scipy 1.10.1, and llvmlite 0.41.1, seven llvmlite releases behind what 3.11+ resolves to. The [dask] extra was already uninstallable on 3.8, since no dask >= 2024.1 supports it.

 - pyproject: requires-python >= 3.11, drop stale classifiers
 - CI: drop the 3.8 jobs; move the special-purpose 3.9/3.10 jobs (--forked, benchmarks, fp32, coverage, notorch, min-deps) to 3.11. The min-deps job moves off the excluded 3.8/macos slot, so the exclude list is now empty.
 - broken_trans_deps: drop the five constraints that only applied to the removed versions (ipython, torch/win-3.10, beartype, mistune, fastjsonschema)
 - requirements: raise floors that cannot install on 3.11. numpy 1.21.0 and scipy 1.7.3 have no cp311 wheels (scipy 1.8.0 caps at <3.11), and torch 1.13.0 ships a cp311 wheel only for manylinux x86_64, which the macos min-deps job needs.
 - ruff: requires-python drives ruff's target-version, so raising it turns on UP007/UP035/UP042/UP045 across the codebase (90 findings). Ignored here and left to a separate modernization pass.

### DIFF
--- a/.github/workflows/pnl-ci-docs.yml
+++ b/.github/workflows/pnl-ci-docs.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
         os: [ubuntu-latest, windows-latest, macos-latest]
         pnl-version: ['head']
         include:
--- a/.github/workflows/pnl-ci.yml
+++ b/.github/workflows/pnl-ci.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
         extra-args: ['']
         os: [ubuntu, macos, windows]
         version-restrict: ['']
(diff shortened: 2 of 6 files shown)


jvesely:
> ruff: requires-python drives ruff's target-version, so raising it turns on UP007/UP035/UP042/UP045 across the codebase (90 findings). Ignored here and left to a separate modernization pass.

I don't think we should do this. If there's a cleanup change needed, it should be merged before the bump rather than after.
afaict, most of the issues should be addressed in #3617.

> CI: drop the 3.8 jobs; move the special-purpose 3.9/3.10 jobs (--forked, benchmarks, fp32, coverage, notorch, min-deps) to 3.11. The min-deps job moves off the excluded 3.8/macos slot, so the exclude list is now empty.

I think it's OK to keep the exclude and drop the next version from the matrix. The distribution was: drop one lower-end job and replace it with a "min-version" special variant. Drop at least 2 middle Python versions and replace them with special test variants.

jvesely:
I've posted an alternative with the above comments at https://github.com/PrincetonUniversity/PsyNeuLink/pull/3621 (only 3.8 and 3.9 dropped). cc @davidt0x

PR #3619, opened 2026-08-07
Hierarchical PEC Fitting #1: KDE+LaplaceEM

aryan-pan:
Fits a group of participants jointly rather than one at a time, so each participant's estimate is informed by the population. Estimation is by empirical-Bayes Laplace EM, with participants drawn from `z_s ~ N(beta, diag(sigma))` in an unconstrained space derived from the search ranges.

```
pec = ParameterEstimationComposition(
    ..., data=stacked, fit_method="hierarchical",
    hierarchical_options={"subject_id": "subject"},
    distributed_options={"pec_factory": build_participant_pec},
)
results = pec.run()
results.group_parameters
```

`hierarchical_options["subject_id"]` in the data identifies participants. PEC requires every column of data to be an outcome variable, so the column is removed both from the group composition's own data and from each participant's slice before it reaches the factory. pec_factory builds one participant's model from their slice, and is required because a Composition cannot be copied.

Setting distributed=True fits participants across a Dask cluster, one per task, resolved through the same helpers as distributed maximum-likelihood fitting. Results are collected by participant index rather than completion order, so distributed and in-process fits agree exactly.

Documented in `HierarchicalFitting.rst`, with a runnable example that fits a simulated group and reports recovery.

### DIFF
--- a/.github/workflows/pnl-ci.yml
+++ b/.github/workflows/pnl-ci.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.10', '3.13', '3.14']
         extra-args: ['']
         os: [ubuntu, macos, windows]
         version-restrict: ['']
@@ -45,12 +45,12 @@ jobs:
         include:
           # --forked run of python only tests
           # Python tests are enough to test potential naming issues
-          - python-version: '3.9'
+          - python-version: '3.11'
             os: ubuntu
             extra-args: '--forked -m "not llvm"'
 
-          # --benchmark-enable run on macos python 3.9
-          - python-version: '3.9'
+          # --benchmark-enable run on macos python 3.11
+          - python-version: '3.11'
             os: macos
             # pytest needs both '--benchmark-only' and '-m benchmark'
             # The former fails the test if benchmarks cannot be enabled
@@ -59,28 +59,29 @@ jobs:
             # https://github.com/ionelmc/pytest-benchmark/issues/243
             extra-args: '-m benchmark --benchmark-enable --benchmark-only --benchmark-min-rounds=2 --benchmark-max-time=0.001 --benchmark-warmup=off -n0 --dist=no'
 
-          # fp32 run on linux python 3.10
-          - python-version: '3.10'
+          # fp32 run on linux python 3.12
+          - python-version: '3.12'
             os: ubuntu
             extra-args: '--fp-precision=fp32'
 
-          # code-coverage run on macos python 3.10
-          - python-version: '3.10'
+          # code-coverage run on macos python 3.12
+          - python-version: '3.12'
             os: macos
             extra-args: '--cov=psyneulink'
 
-          # run without pytorch on windows python 3.10
-          - python-version: '3.10'
+          # run without pytorch on windows python 3.12
+          - python-version: '3.12'
             os: windows
             torch: 'notorch'
 
-          # add python 3.8 with deps restricted to min supported package versions
-          - python-version: '3.8'
+          # add python 3.10 with deps restricted to min supported package versions
+          - python-version: '3.10'
             os: macos
             version-restrict: 'min'
 
         exclude:
-          - python-version: '3.8'
+          # Replaced with the min version run above
+          - python-version: '3.10'
             os: macos
 
     steps:
--- a/psyneulink/core/compositions/hierarchical/distributedestep.py
+++ b/psyneulink/core/compositions/hierarchical/distributedestep.py
@@ -1,0 +1,194 @@
+# Princeton University licenses this file to You under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may obtain a copy of the License at:
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+
+
+# ************************************  Distributed E-step  ***********************************************************
+
+"""E-step distributed across a Dask cluster.
+
+One task per participant: build that participant's model on a worker, fit it, return the posterior
+summary.  The M-step stays on the driver, so this runner is interchangeable with the in-process one.
+
+Building a participant's model constructs a composition and compiles it, so each worker caches the
+models it builds.  That only pays off if a participant keeps landing on the same worker, hence the
+pinning after the first iteration; otherwise every worker ends up holding a model for every
+participant, which is how a large fit exhausts memory.
+
+Dask is imported inside the functions that need it, so this module can be imported, and its worker
+task exercised, without it.
+"""
+
+import uuid
+
+import numpy as np
+
+from psyneulink.core.components.functions.nonstateful import fitfunctions as _fitfunctions
+from psyneulink.core.compositions.hierarchical.laplaceem import (
+    EStepConfig,
+    EStepResult,
+    log_gauss_diag,
+    subject_laplace_objective,
+    subject_map_estep,
+)
+from psyneulink.core.compositions.hierarchical.transforms import BoundedTransform
+
+__all__ = ["make_distributed_estep_runner"]
+
+#: Per-worker cache of participant models, used when no Dask worker context is available.  Keeping
+#: it module-level lets the task body be exercised in-process, without a cluster.
+_SUBJECT_FALLBACK_CACHE = {}
+
+
+def _worker_subject_cache(fit_id):
+    """Return the cache dict for this worker, creating it if needed."""
+    try:
+        from dask.distributed import get_worker
+        worker = get_worker()
+    except (ImportError, ValueError):
+        # No worker context: either Dask is absent, or this is running on the driver.
+        return _SUBJECT_FALLBACK_CACHE
+    cache = getattr(worker, "_hierarchical_subject_cache", None)
+    if cache is None:
+        cache = worker._hierarchical_subject_cache = {}
+    return cache
+
+
+def _dask_subject_estep(
+    pec_factory, subject_index, data_slice, mu_s, sigma, lower, upper, z0, worker_cores,
+    fit_id, config,
+):
+    """Fit one participant on a worker and return their posterior summary.
+
+    The whole call holds the evaluation lock, not just the model construction: two compiled models
+    driven at once in a single process is what the lock exists to prevent.
+    """
+    with _fitfunctions._PEC_EVALUATION_LOCK:
+        cache = _worker_subject_cache(fit_id)
+        key = (fit_id, subject_index)
+        if key not in cache:
+            from psyneulink.core.globals.threads import set_num_threads
+            if worker_cores is not None:
+                set_num_threads(worker_cores)
+            cache[key] = pec_factory(data_slice, subject_index)
+        pec, inputs = cache[key]
+
+        transform = BoundedTransform(lower=lower, upper=upper)
+        sigma = np.asarray(sigma, dtype=float)
+        mu_s = np.asarray(mu_s, dtype=float)
+
+        def neg_log_post(z):
+            theta = transform.to_natural(z)
+            return -float(pec.log_likelihood(*theta, inputs=inputs)) - log_gauss_diag(z, mu_s, sigma)
+
+        post = subject_map_estep(neg_log_post, z0=z0, prior_variance=sigma, config=config)
+
+        worker_address = None
+        try:
+            from dask.distributed import get_worker
+            worker_address = get_worker().address
+        except (ImportError, ValueError):
+            pass
+
+        return subject_index, post, worker_address
+
+
+def make_distributed_estep_runner(
+    client, pec_factory, data_slices, bounds, *, config=None, worker_cores=None, fit_id=None,
+):
+    """Build an E-step that fits every participant at once, across a cluster.
+
+    Arguments
+    ---------
+
+    client : dask.distributed.Client
+        Cluster to submit to.
+
+    pec_factory : callable
+        ``pec_factory(data, subject_index=None) -> (pec, inputs)``.  Must be importable on the
+        workers, so it has to be defined at module level rather than nested or bound.
+
+    data_slices : sequence of pandas.DataFrame
+        One participant's trials each, in participant order.  Only the relevant slice is sent to
+        each task.
+
+    bounds : tuple
+        ``(lower, upper)`` arrays of the search range, shared by every participant.
+
+    config : EStepConfig : default None
+        Settings for each participant's optimization.
+
+    worker_cores : int : default None
+        Threads to give the compiled model on a worker.
+
+    fit_id : str : default None
+        Identifies this fit in the per-worker model cache.  Reusing a value across calls against the
+        same cluster lets the workers keep the models they already built.
+
+    Returns
+    -------
+
+    A callable ``runner(mu, sigma, prev_z, warm_start) -> EStepResult``, interchangeable with the
+    in-process runner.
+    """
+    config = config if config is not None else EStepConfig()
+    fit_id = fit_id or uuid.uuid4().hex
+    lower, upper = (np.asarray(b, dtype=float) for b in bounds)
+    data_slices = list(data_slices)
+    # Filled in after the first iteration, so each participant returns to the worker holding
+    # their model instead of being rebuilt somewhere else.
+    home = {}
+
+    def runner(mu, sigma, prev_z, warm_start):
+        n_subjects, n_params = mu.shape
+        futures = []
+        for s in range(n_subjects):
+            z0 = np.asarray(prev_z[s] if warm_start else mu[s], dtype=float)
+            submit_kwargs = {"pure": False}
+            if s in home:
+                # allow_other_workers so a dead worker reschedules rather than hanging; the
+                # participant re-pins on the next iteration.
+                submit_kwargs.update(workers=[home[s]], allow_other_workers=True)
+            futures.append(client.submit(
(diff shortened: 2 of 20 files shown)


psyneulink/core/compositions/hierarchical/subjectlikelihood.py
```diff
@@ -0,0 +1,316 @@
...
+    "SubjectSplit",
+    "split_stacked_data",
+]
+
+#: Participant count above which building every model in one process is likely to exhaust memory.
+#: Each participant's model is compiled separately, so they accumulate.
+IN_PROCESS_SUBJECT_WARN_THRESHOLD = 32
+
+
+def _comparable_names(qualified):
```
davidt0x, line 50: Makes sense to strip the instance names here, since `DDM-6.rate` and `DDM-7.rate` can be the same parameter across participant models. One edge case, though: suppose a model has two DDM mechanisms named `left` and `right`, and we fit `rate` on both. Both names become `"rate"`, so this check won’t catch the factory returning those parameters in the opposite order for another participant, assuming their bounds match. That could happen accidentally if the parameter dictionary is built by iterating over a set, since its order can differ across worker processes.

Probably fine to rely on the factory’s ordering contract for now, but could we mention that limitation here and recommend constructing the parameters in a consistent order?

psyneulink/core/compositions/hierarchical/distributedestep.py
```diff
@@ -0,0 +1,241 @@
...
+
+        z_hat = np.empty((n_subjects, n_params))
+        variance = np.empty((n_subjects, n_params))
+        curvature = np.empty((n_subjects, n_params))
+        steps = np.empty((n_subjects, n_params))
+        subject_objective = np.empty(n_subjects)
+        success = np.empty(n_subjects, dtype=bool)
+        messages = []
+
+        for subject_index, post, worker_address in client.gather(futures):
```
davidt0x, line 211: Could cleanup run before the other participant tasks finish here? If one task raises, `gather()` propagates the error while another worker might still be building its model. That worker could then add its model to the cache after `release()` has cleared it, leaving it in memory on a caller-owned cluster.

Maybe we could call `wait(futures, return_when="ALL_COMPLETED")` before `gather()` so cleanup happens after all tasks finish. That would delay reporting the error until the remaining tasks complete, but seems like a straightforward fix. A two-worker test where one task fails while the other is still building would be useful too.

psyneulink/core/compositions/hierarchical/laplaceem.py
```diff
@@ -0,0 +1,523 @@
...
+        simplex = np.vstack([z0] + [z0 + SIMPLEX_SCALE * e for e in np.eye(n)])
+        options = {"xatol": 1e-6, "fatol": 1e-6, "maxiter": 2000, "initial_simplex": simplex}
+    if config.optimizer_options:
+        options.update(config.optimizer_options)
+
+    result = minimize(neg_log_post, z0, method=config.method, options=options)
+    z_hat = np.asarray(result.x, dtype=float)
+
+    step = config.resolve_hessian_step(prior_variance)
+    f0 = float(result.fun)
```
davidt0x, line 350: Could we check that result.fun is finite before computing curvature here? Falling back to the prior variance makes sense when a finite-difference probe lands outside the support, but if the objective at the fitted point is itself infinite, we don’t have a valid posterior.
With a toy likelihood returning -inf, the fallback leaves the group parameters unchanged, so the fit returns converged=True with objective=-inf, even though all participant fits failed. There is a warning, but it might be clearer to stop with an error identifying the participant whose objective was invalid. A small regression test for this would be useful too.

.github/workflows/pnl-ci.yml
```diff
@@ -94,6 +95,13 @@ jobs:
             arch: 'default'
             version-restrict: 'min'
 
+          # parameter estimation tests with optional dependencies on linux python 3.10
+          - python-version: '3.10'
+            os: ubuntu
```
davidt0x: Why not enable these tests on all OS?

psyneulink/core/compositions/hierarchical/laplaceem.py
```diff
@@ -0,0 +1,523 @@
...
+covariance.  The M-step updates `beta` by least squares on the modes and `sigma` from the posterior
+second moments.
+
+Only the diagonal of the curvature is computed, matching the diagonal group covariance, so a
+participant's reported variance is ``1 / H_kk`` rather than ``(H^-1)_kk``: the variance of one
+parameter with the others held at the mode, not with them integrated out.  Where parameters trade
+off against each other the first is the smaller, so intervals err narrow.
+
+Curvature comes from central finite differences, which requires the objective to be deterministic in
+`theta`.  Participant models must therefore be built with common random numbers; without them the
```
davidt0x: Do we enforce this? We should throw an error if a user constructs the PEC without common random numbers.

psyneulink/core/compositions/hierarchical/laplaceem.py
```diff
@@ -0,0 +1,523 @@
...
+        step = np.where(unusable, 0.5 * step, step)
+        curvature = np.where(
+            unusable, diagonal_hessian(neg_log_post, z_hat, step=step, f0=f0), curvature
+        )
+
+    # Curvature that is zero or negative describes a fit that is flat or rises away from the
+    # mode, and one that is still infinite describes a probe that never fit. Neither is a width,
+    # so the prior stands in and reports that this parameter was not pinned down.
+    usable = np.isfinite(curvature) & (curvature > 0)
+    inverse = np.where(usable, 1.0 / np.where(usable, curvature, 1.0), prior_variance)
```
davidt0x: Codex caught this one:

I see the diagonal approximation is documented, but these variances also feed into the group variance update, so the effect goes beyond narrower uncertainty intervals. When parameters trade off, 1 / H[k,k] can be quite different from (H⁻¹)[k,k]. A small Gaussian test with H = [[10, 9], [9, 10]] gives variances of 0.1 here versus about 0.526 for the exact posterior.
Could we consider computing the full participant Hessian while keeping the group covariance diagonal? If we want to keep this approximation for speed, it would be useful to add a correlated-likelihood test and check how much it affects parameter recovery in the DDM example.

davidt0x: This will of course increase the number of likelihood evaluations as number of parameters increase. However, I think we will be able to mitigate that with the GPU stuff I have been working on.

davidt0x: Here is a example test illustrating the issue from codex:

```python
"""Review reproduction: compare diagonal curvature with an exact Gaussian posterior.

The independent case should pass. The coupled case intentionally fails on PR #3619:
the approximation reports conditional variances instead of marginal variances.
"""

import numpy as np
import pytest

from psyneulink.core.compositions.hierarchical.laplaceem import subject_map_estep


@pytest.mark.parametrize("coupled", [False, True], ids=["independent", "parameter_tradeoff"])
def test_estep_variance_matches_exact_gaussian_posterior(coupled):
    # Independent standard-normal priors on z[0] and z[1].
    # Observation noise variance is 1/9, with observed values equal to zero.
    # Independent case: observe each parameter separately.
    # Coupled case: observe only their sum, allowing them to trade off.
    def neg_log_post(z):
        prior = 0.5 * np.dot(z, z)
        residual_squared = np.sum(z) ** 2 if coupled else np.dot(z, z)
        return float(prior + 0.5 * 9.0 * residual_squared)

    post = subject_map_estep(
        neg_log_post, z0=np.zeros(2), prior_variance=np.ones(2)
    )

    # These are exactly quadratic objectives, so neither simulation noise nor
    # finite-difference truncation error explains a covariance discrepancy.
    assert post.success
    np.testing.assert_allclose(post.z_hat, np.zeros(2), atol=1e-6)
    np.testing.assert_allclose(post.curvature, [10.0, 10.0], rtol=1e-6)

    # Independent: H = 10 I, so diag(inv(H)) = [0.1, 0.1].
    # Coupled: H = [[10, 9], [9, 10]], whose determinant is 19;
    # inv(H) = [[10, -9], [-9, 10]] / 19.
    exact_variance = np.full(2, 10.0 / 19.0 if coupled else 0.1)
    np.testing.assert_allclose(post.variance, exact_variance, rtol=1e-6)
```

Scripts/Examples/ParameterEstimation/hierarchical/hierarchical_fitting.py
```diff
@@ -0,0 +1,179 @@
+"""Fit a group of participants jointly with `ParameterEstimationComposition`.
```
davidt0x, line 1: Since the docs link to this as the complete user example, could we move these files into Scripts/Examples/ParameterEstimation/hierarchical/ and update the documentation links? That would make them easier to discover and leave room for other parameter-estimation examples alongside this one.

Something like:

Scripts/Examples/ParameterEstimation/hierarchical/
    hierarchical_fitting.py
    make_example_data.py
    submit_hierarchical.slurm

Scripts/Debug/pec_hierarchical/hierarchical_fitting.py
```diff
@@ -0,0 +1,179 @@
...
+    data = load_data(args.data)
+    print(f"{len(data)} trials from {data[args.subject_id].nunique()} participants")
+
+    distributed_options = {"pec_factory": participant_pec}
+    if args.n_workers is not None:
+        distributed_options["n_workers"] = args.n_workers
+
+    # The model given here declares which parameters are fitted and which outputs are compared
+    # What is fitted, and over what range, is declared once, by the factory below.
+    pec = pnl.ParameterEstimationComposition(
```
davidt0x: Could we name this group_fit and use participant_pec for the instance inside the factory? A short comment explaining that the group fit runs EM and calls each participant PEC’s log_likelihood() would help make the two roles clearer. The comment above this also looks outdated: there isn’t a model passed here, and the factory is defined above rather than below.

Scripts/Examples/ParameterEstimation/hierarchical/hierarchical_fitting.py
```diff
@@ -0,0 +1,179 @@
...
+        nodes=[comp],
+        parameters={
+            (name, decision): np.linspace(*FIT_RANGES[name], 1000) for name in FIT_PARAMS
+        },
+        outcome_variables=[
+            decision.output_ports[pnl.DECISION_OUTCOME],
+            decision.output_ports[pnl.RESPONSE_TIME],
+        ],
+        data=data,
+        optimization_function=PECOptimizationFunction(
```
davidt0x, line 100: One possible improvement to try: could we let a participant PEC omit optimization_function when it’s only being used for log_likelihood()? We’d still create a default PECOptimizationFunction internally for the simulation and likelihood machinery, but mark the PEC as likelihood-only and raise a helpful error if someone calls run().

That would let us remove the differential-evolution settings here, which suggest a search is being performed when it isn’t. We’d just need to keep this distinct from the hierarchical group PEC, whose run() still needs to work.

Longer term, it probably makes sense to separate simulation-based likelihood evaluation from PEC entirely, so both ordinary and hierarchical fitting can use it without constructing a PEC just to score parameters. That could be a follow-up rather than something we need to tackle in this PR. Need to talk with Jon about this though.

psyneulink/core/compositions/hierarchical/hierarchicalresults.py
```diff
@@ -0,0 +1,223 @@
...
+            },
+            index=pd.Index(names, name="parameter"),
+        )
+
+        subject_parameters = pd.DataFrame(
+            theta_hat, index=pd.Index(labels, name="subject"), columns=list(names)
+        )
+
+        posteriors = pd.DataFrame({
+            "subject": np.repeat(labels, len(names)),
```
davidt0x: Another catch by codex:

np.repeat(labels, len(names)) converts the labels to a common NumPy type. With participant IDs [1, "1"], the fit correctly treats them as distinct participants. However:
subject_labels:              (1, "1")
subject_parameters.index:    [1, "1"]
subject_posteriors.subject:  ["1", "1"]

That makes posterior results ambiguous when grouping or joining them back to participant data.
Maybe preserve the labels as objects when repeating them. This is an edge case involving mixed ID types, but the resulting corruption is silent.

davidt0x (changes requested):

.github/workflows/pnl-ci.yml
aryan-pan: added CI for windows/macOS as well. if you meant running these by default across all builds, I can make that change

.github/workflows/pnl-ci.yml
```diff
@@ -163,7 +163,7 @@ jobs:
       uses: ./.github/actions/install-pnl
       id: install
       with:
-        features: 'dev'
+        features: 'dev,dask'
```
kmantel, line 166: If this will be an optional component, we should mark all tests that require dask, and don't install dask in the main jobs, skipping those tests, to make sure everything runs fine without it. Then, add a job that installs this feature and includes the dask tests. We do something similar with torch in reverse; it's a main dependency, but we have a separate job to test without it.

psyneulink/core/compositions/hierarchical/laplaceem.py
```diff
@@ -0,0 +1,666 @@
...
+            subject_objective=subject_objective,
+            success=success,
+            messages=tuple(messages),
+        )
+
+    return runner
+
+
+@dataclass
+class LaplaceEMResult:
```
kmantel, line 505: There are several shared attributes with HierarchicalPECResults and EStepResult. Are they different subtypes of a similar thing?

psyneulink/core/compositions/parameterestimationcomposition.py
```diff
@@ -599,6 +634,8 @@ def __init__(
         context: Optional[Context] = None,
         distributed: bool = False,
         distributed_options: Optional[Mapping] = None,
+        fit_method: Optional[Literal["hierarchical"]] = None,
```
kmantel, line 637: This and similar params that have a limited number of options should use an enum/similar, maybe PNLStrEnum

psyneulink/core/compositions/parameterestimationcomposition.py
```diff
@@ -774,6 +877,144 @@ def __init__(
         # this to avoid infinite recursion.
         self._run_called = False
 
+    #: Settings accepted by `hierarchical_options`, with their defaults.
+    _HIERARCHICAL_OPTION_DEFAULTS = {
```
kmantel, line 881: can/should these be `Parameter`s? @davidt0x

tests/composition/pec/test_pec_hierarchical.py
```diff
@@ -0,0 +1,1332 @@
...
+
+def test_dtheta_dz_matches_numerical():
+    t = BoundedTransform(lower=[0.0, 1.0], upper=[2.0, 5.0])
+    z = np.array([0.3, -0.7])
+    h = 1e-6
+    num = np.array([
+        (t.to_natural(z + h * e)[k] - t.to_natural(z - h * e)[k]) / (2 * h)
+        for k, e in enumerate(np.eye(2))
+    ])
+    assert np.allclose(t.dtheta_dz(z), num, rtol=1e-5)
```
kmantel, line 155: use `np.testing.assert_allclose`, because it also checks shape, here and for others

kmantel (changes requested):
There's a lot about this/PEC that I don't know the motivations behind, so I'll defer more to Dave; I wonder about how much this is meant as a somewhat independent library-extension/plugin vs something more tightly integrated with and designed like the rest of PNL.

.github/workflows/pnl-ci.yml
davidt0x: Yeah, this is my bad. Aryan had it setup like this at first and I told him to change it. The more I think about it I think you and him are right, it should optional and tests in a smaller separate test job.

psyneulink/core/compositions/hierarchical/laplaceem.py
aryan-pan: Thanks for the comments! `LaplaceEMResult` is the EM fitting routine’s numerical output. `HierarchicalPECResults` keeps those same arrays and adds user-facing tables with parameter names and values in the model’s original units. `EStepResult` is an intermediate result used within the fitting loop.

psyneulink/core/compositions/hierarchical/laplaceem.py
davidt0x: Should we make the Full-Hessian inversion the default?

Scripts/Examples/ParameterEstimation/hierarchical/hierarchical_fitting.py
davidt0x: Did you try addressing this one?

tests/composition/pec/test_pec_hierarchical.py
```diff
@@ -0,0 +1,1571 @@
...
+def test_provider_reads_names_and_bounds_from_the_model():
+    frames = split_stacked_data(_stacked_frame(), "subject").frames
+    factory = lambda data, subject_index=None: (  # noqa: E731
+        _StubPEC(["rate", "threshold"], [(-1.5, 1.5), (0.3, 1.5)]), None
+    )
+    provider = PECFactorySubjectLikelihood(factory, frames)
+    assert provider.fit_param_names == ("rate", "threshold")
+    assert provider.n_params == 2 and provider.n_subjects == 3
+    lower, upper = provider.bounds
+    np.testing.assert_allclose(lower, [-1.5, 0.3]) and np.allclose(upper, [1.5, 1.5])
```
davidt0x: I don't think this works. Since assert_allclose returns None, the upper-bound check never executes. This needs two separate assertions.

psyneulink/core/compositions/hierarchical/laplaceem.py
aryan-pan: Yes, made it the default now

Scripts/Examples/ParameterEstimation/hierarchical/hierarchical_fitting.py
aryan-pan: `optimization_function` is now optional if `data` is provided. And yeah, decoupling completely does warrant another PR imo

davidt0x (approved):
LGTM

PR #3627, opened 2026-08-20
Drop pillow from the requirements.

jvesely:
Not used anywhere.


kmantel:
I was surprised this didn't break, because we have PIL in top-level imports (composition.py, showgraph.py), and uninstalling pillow does so. But it looks like it's a dependency of one of the other packages. I think we should probably still keep it unless we want to make showgraph stuff optional.

jvesely:
> I was surprised this didn't break, because we have PIL in top-level imports (composition.py, showgraph.py), and uninstalling pillow does so. But it looks like it's a dependency of one of the other packages. I think we should probably still keep it unless we want to make showgraph stuff optional.

my bad. Only checked for "pillow".

PR #3632, opened 2026-08-29
Support >2D input/output for components

kmantel:
Projection/MatrixTransform can be provided matrices of >2D, using numpy
tensordot. If input/output to a MatrixTransform is >2D but matrix is 2D, the
transform is applied in batch, according to numpy's default behavior.

- components:
	- Provide new/replacement methods to help get and check valid input values:
		- `default_external_input`: returns an array that would be used as input
		to the Component when executed if no input is passed in
		- `external_input_shape`: returns the shape of a single valid input to
		the Component
		- `parse_input_array`: given an array-like object, uses that object
		(modified, if necessary) to create and return a valid input or list of
		multiple inputs to the Component

- Port:
	- Reshape default_input to default value shape. The result of the Port's
	function may have a different shape than the input, so returning the
	default variable directly must also conform to the output shape

- Utilities:
	- Add equivalents of several numpy functions, with awareness of the
	shapes of nested arrays in ragged arrays:
		- `shape`
		- `clip`
		- `full`
		- `full_like`
		- `zeros`
		- `zeros_like`
		- `ones`
		- `ones_like`

Currently, these components remain limited to 2D in some way:
- AutoAssociativeProjection
    - matrix restricted to 2D (uses Hollow Matrix)
- ContentAddressableMemory
- ContrastiveHebbian
- ContrastiveHebbianMechanism
- EMComposition
- EMStorageMechanism
- Hebbian
- KohonenLearningMechanism
    - matrix restricted to 2D (uses Hollow Matrix)
- LeabraFunction
- LeabraMechanism
- Stability
    - matrix restricted to 2D (uses Hollow Matrix)


psyneulink/core/globals/utilities.py
```diff
@@ -2657,3 +2681,464 @@ def get_stacklevel_skip_file_prefixes(
...
+        except ValueError as e:
+            if 'The truth value of an array with more than one element is ambiguous' in str(e):
+                return [_clip(x_i) for x_i in x]
+            else:
+                raise
+    return convert_all_elements_to_np_array(_clip(convert_all_elements_to_np_array(a)))
+
+
+# ----------------------------------------------------------------------
+# _validate_np_tensordot_args is based heavily on contents of
```
kmantel, line 3056: I think this is the right attribution

kmantel:
The smaller commits should all stand on their own

PR #3641, opened 2026-09-07
treewide: Prefer ravel() instead of flatten()

jvesely:
Unless flatten()'s copying semantics are required.


kmantel (approved):
I wasn't aware of the difference. I agree we should use ravel.

PR #3647, opened 2026-09-12
ci: test-release: drop python 3.9 like pnl-ci

kmantel:


kmantel:
I wonder if we could/should do something to use the same matrix in pnl-ci.yml and test-release.yml. There's a related discussion: https://github.com/orgs/community/discussions/26284

jvesely:
ah, sorry about that. I missed this when dropping the old Pythons.

> I wonder if we could/should do something to use the same matrix in pnl-ci.yml and test-release.yml. There's a related discussion: https://github.com/orgs/community/discussions/26284

The matrices are different, though. test-release doesn't have any of the extra jobs and doesn't have the special "runs-on" parsing. We can sync that, but I don't think the runs-on part can be easily reused, which means the 'arch' part of the matrix would need to be dropped. It might be easier to just have a comment to check test-release when adding/dropping python versions.

kmantel:
The difference between the two is why I brought it up really, because it's not testing the same combos. But if that's not unintentional, then I don't think it's a big deal.

jvesely:
> The difference between the two is why I brought it up really, because it's not testing the same combos. But if that's not unintentional, then I don't think it's a big deal.

A bit more background: The idea was that since the regular job runs for the tagged commit anyway, we don't need to retest the same jobs, especially the jobs targeting corner-case functionality. Rather, we should test the sanity and completeness of the uploaded install packages (.sdist/.whl).
There might be a way for the test-release workflow to fail if the regular pnl-ci workflow failed for the same commit id.

We could also potentially merge the two jobs and use `if:` to prune steps that aren't needed in one or the other. However, most of the steps are different (installation, even test execution), and there would be little overlap with a significant increase in complexity.

